Issue #7, opened 2020-03-20
web: split web ui and api contianers


ppanero:
Some nginx configuration is required. See [here](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/docker/nginx/conf.d/default.conf)

Issue #6, opened 2020-03-20
nginx: consolidate configuration


ppanero:
The current configuration is loaded from a ConfigMap, which overwrites the one done in `cookiecutter-invenio-rdm`. Therefore many of the features are not being used (e.g. off-loading of statics).

Refs:

- [Load conf.d modules](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/docker/nginx/nginx.conf#L76)
- [Static off-load and others](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/docker/nginx/conf.d/default.conf#L120)

PR #8, opened 2020-03-20
web/nginx: shared volume (pvc - persistent volume claim)

ppanero:
Now Nginx and Web nodes share a volume (Persistent) for statics. However, it needs the image to have the statics (Addresse here https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/56).

Since volumes override the folder where they are being mounted (empty them). The web pod folder is "wiped out" of statics. To solve the issue an `initContainer` is launched, this will mount the volume in a temporary location and copy the statics to it. Therefore, it will be initialized with the correct statics. The copy of those is made to overwrite old ones, so there is no problem with future deployments.

The serving of statics has been off-loaded to nginx. Deeper consolidations are out of the scope of this PR, but will be addressed as part of https://github.com/inveniosoftware/helm-invenio/issues/7 and https://github.com/inveniosoftware/helm-invenio/issues/6.

In addition, at the moment the `initContainer` is launched once per pod. Which means that if there are 6 web pods the assets will be copied 6 times. This is not an issue since OpenShift provides a distributed locking system and all pod's assets are the same. This will be improved by adding nginx as a sidecar container and using an `emptyDir` instead of a volume. Will be treated in https://github.com/inveniosoftware/helm-invenio/issues/9 

Requires: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/56

### DIFF
--- a/invenio/templates/application.yaml
+++ b/invenio/templates/application.yaml
@@ -44,10 +44,10 @@ spec:
         command: [
           '/bin/bash',
           '-c',
-          'cp -R {{ .Values.web.statics.location }}/. /tmp/invenio/statics/']
+          'cp -R {{ .Values.web.assets.location }}/. /tmp/invenio/assets/']
         volumeMounts:
-            - name: "invenio-statics"
-              mountPath: "/tmp/invenio/statics/"
+            - name: "invenio-assets"
+              mountPath: "/tmp/invenio/assets/"
       containers:
       - name: web
         image: {{ .Values.web.image }}
@@ -112,8 +112,8 @@ spec:
         volumeMounts:
           - name: uwsgi-config
             mountPath: '/opt/invenio/src/uwsgi'
-          - name: "invenio-statics"
-            mountPath: "{{ .Values.web.statics.location }}"
+          - name: "invenio-assets"
+            mountPath: "{{ .Values.web.assets.location }}"
         resources:
           requests:
             memory: 200Mi
@@ -124,9 +124,9 @@ spec:
           configMap:
             defaultMode: 420
             name: uwsgi-config
-        - name: "invenio-statics"
+        - name: "invenio-assets"
           persistentVolumeClaim:
-            claimName: "invenio-statics-claim"
+            claimName: "invenio-assets-claim"
 
 ---
 apiVersion: apps.openshift.io/v1
--- a/invenio/templates/application.yaml
+++ b/invenio/templates/application.yaml
@@ -38,6 +38,16 @@ spec:
       labels:
         app: web
     spec:
+      initContainers:
+      - name: web-init
+        image: {{ .Values.web.image }}
+        command: [
+          '/bin/bash',
+          '-c',
+          'cp -R {{ .Values.web.statics.location }}/. /tmp/invenio/statics/']
+        volumeMounts:
+            - name: "invenio-statics"
+              mountPath: "/tmp/invenio/statics/"
       containers:
       - name: web
         image: {{ .Values.web.image }}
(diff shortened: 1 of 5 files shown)


invenio/templates/application.yaml
```diff
@@ -38,6 +38,16 @@ spec:
...
     spec:
+      initContainers:
+      - name: web-init
+        image: {{ .Values.web.image }}
+        command: [
+          '/bin/bash',
+          '-c',
+          'cp -R {{ .Values.web.statics.location }}/. /tmp/invenio/statics/']
+        volumeMounts:
+            - name: "invenio-statics"
```
Glignos: This might just be me but I don't like the word 'statics', just nitpicking here.

invenio/templates/application.yaml
```diff
@@ -38,6 +38,16 @@ spec:
       labels:
         app: web
     spec:
+      initContainers:
```
Glignos: I was thinking that maybe the initContainer  could be used to initialize the NGINX Pod instead since there will be always fewer Nginx-Pods and maybe moving it there, you could remove the need for using a persistent volume and just copy it to a local volume of this pod(unsure if you're going to have the same mounting-wiping issue.

invenio/templates/volumes.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: "v1"
+kind: "PersistentVolumeClaim"
+metadata:
+  name: "invenio-statics-claim"
+spec:
+  accessModes:
+    - "ReadWriteMany"
+  resources:
+    requests:
+      storage: "10Gi"
```
Glignos: Maybe this could prove to be a bit unelastic, too big or too small. Should this also be a Helm value?

ppanero: Done :)

invenio/templates/application.yaml
ppanero: Discussed IRL, to be seen after testing nginx as sidecar.

invenio/templates/application.yaml
ppanero: renamed to assets

invenio/templates/application.yaml
```diff
@@ -102,6 +112,8 @@ spec:
         volumeMounts:
           - name: uwsgi-config
             mountPath: '/opt/invenio/src/uwsgi'
+          - name: "invenio-assets"
```
zzacharo: @ppanero Sorry for the noob question: The web pod needs to mount the "invenio-assets" volume in the web pod for what? I understand that you need to provide the volume so the `initContainer` can copy files in it right? But the web container needs to mount it too for which reason?

Also I assume that as long as  the `invenio-assets` volume is actually the "invenio-assets-claim" persistent volume, the copy that is happening in the `initContainer` will persist the data there?
The persistent volume is created/mounted only once or for every web pod and in the end it persists the data the last pod deposited into it?

invenio/templates/volumes.yaml
zzacharo: @ppanero sorry how is done? I still see the `10Gi` value hardcoded here...is github messing with me???

invenio/templates/application.yaml
zzacharo: Discussed IRL

ppanero: Discussion point: Yes, Zach is right. However, since Nginx might become a sidecar (CC @Glignos) there is no point on changing this now, let's see the outcome of the sidecar work :)

PR #11, opened 2020-03-25
Added pure Kubernetes helm chart

wgresshoff:
As first step I added a specific directory for pur K8s helm chart.


fenekku:
@wgresshoff @ppanero How is this coming along? Should it be moved to the May board? Just checking :)

wgresshoff:
From my point of view: yes, it should be moved. We should integrate it into one helm script which is working on Openshift and Kubernetes as well. And I removed the livenessProbes to get it working in Kubernetes. Don't have more spare time in April to work on that.

ppanero:
@fenekku I fully agree with @wgresshoff. As is, it is working in bare K8 which is an awesome work. Now we should merge it in just one chart as Werner said. It will come in May or June (I will also work on it, later on). Let's move it :)

origliante:
@fenekku @lnielsen hey guys, is there some testing I could do on this, to help move forward the merge?

jlebonzec (approved):
I reviewed @wgresshoff and I believe it's very nice to start with. Thanks for this work!

On my own cluster, I managed to get a running chart after creating secrets and playing around the Ingress configuration. It showed some DOM Exceptions when trying to upload, but I believe this is a great starting point.

It's not modifying any existing OpenShift chart, so we don't risk breaking anything there either.

I will therefore merge and provide changes through additional PR.

PR #15, opened 2020-07-23
celery: add beat deployment config

ntarocco:
Warning: I have not tested it! :)


topless:
I fixed a couple of issues and tested it, worker-beat container works fine (tested it in ILS) but it does not have health checks to ensure it is running correctly. In contrast to celery worker celery beat does not support celery inspect ping. We have to come up with another way to do the health check for celery beat.

ppanero (approved):
Fixed the command and parametrized a few things. Tested in zenodo-rdm, seems to work!

ppanero:
Changes applied, works for now. De-duplication might need to be part of https://github.com/inveniosoftware/helm-invenio/issues/24

PR #33, opened 2021-12-13
pvc: add web app shared volume

ppanero:


ppanero:
from @wgresshoff 
>  `volume.beta.kubernetes.io/storage-class` and `volume.alpha.kubernetes.io/storage-class` in `metadata.annotations`. These are deprecated and should be renamed to `storageClassName` instead and be moved to `spec`.

invenio/templates/pvc.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.persistence.enabled }}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ .Values.persistence.name }}
+  labels:
+    app: web
+spec:
+  storageClassName: {{ .Values.persistence.storage_class }}
```
ppanero, line 9: @wgresshoff fixed! Thanks for the review. I checked and if the default value of the variables is `""` then it takes the default storge class configured in the cluster, so we are good :) no need for `if`'s

PR #39, opened 2022-01-24
Changed the charts at some places to better meet (local) K8s requirements

wgresshoff:
- make externally defined pvc usable, so helm won't delete it when invenio is uninstalled
- add resource definitions to values.yaml to make them configurable
- fix datacite secret
- add possibility to NOT define an Ingress (as in our K8s we are using Istio instead of nginx)
- fix imagePullSecrets in web.yaml (it was under spec->template->spec->containers)

### DIFF
--- a/invenio/templates/deployments/routing.yaml
+++ b/invenio/templates/deployments/routing.yaml
@@ -15,6 +15,7 @@ spec:
     insecureEdgeTerminationPolicy: Redirect
     termination: edge
 {{- else }}
+{{- if .Values.ingress.enabled }}
 {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
 apiVersion: networking.k8s.io/v1
 {{- else }}
@@ -50,4 +51,5 @@ spec:
             backend:
               serviceName: {{ if .Values.haproxy.enabled }}haproxy{{ else }}web{{ end }}
               servicePort: http
+{{- end }}
 {{- end }}
--- a/invenio/templates/deployments/web.yaml
+++ b/invenio/templates/deployments/web.yaml
@@ -117,6 +117,10 @@ spec:
           mountPath: '/opt/invenio/src/uwsgi'
         - name: nginx-invenio-assets
           mountPath: /opt/nginx-invenio-assets
+        {{- if .Values.persistence.enabled }}
+        - mountPath: /opt/invenio/var/instance/data
+          name: shared-volume
+        {{- end }}
         resources:
           requests:
               cpu: 500m
@@ -185,6 +189,7 @@ spec:
       {{- end }}
       - name: nginx-config
         configMap:
+          defaultMode: 420
           name: nginx-config
       - name: var-run
         emptyDir: {}
@@ -198,3 +203,8 @@ spec:
           name: uwsgi-config
       - name: nginx-invenio-assets
         emptyDir: {}
+      {{- if .Values.persistence.enabled }}
+      - name: shared-volume
+        persistentVolumeClaim:
+          claimName: {{ .Values.persistence.name }}
+      {{- end }}
--- a/invenio/templates/deployments/web.yaml
+++ b/invenio/templates/deployments/web.yaml
@@ -90,6 +90,9 @@ spec:
               key: {{ default (printf "INVENIO_%s_APP_CREDENTIALS" .name) }}
         {{- end }}
         {{- end }}
+        {{- if .Values.extraEnv }}
+        {{- toYaml .Values.extraEnv | nindent 8 }}
+        {{- end }}
         readinessProbe:
           exec:
             command:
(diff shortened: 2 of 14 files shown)


invenio/templates/deployments/routing.yaml
```diff
@@ -15,6 +15,7 @@ spec:
     insecureEdgeTerminationPolicy: Redirect
     termination: edge
 {{- else }}
+{{- if .Values.ingress.enabled }}
```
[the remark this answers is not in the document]
wgresshoff: Use case is: our new K8s cluster uses the Istio ingress instead of nginx-ingress. So I think it's better to make the ingress excludable until there is a better solution (perhaps defining it in values.yaml and using toYaml to include it in the chart). 
As OpenShift doesn't have the needs for an Ingress (there is Route instead) I thought the if is better placed around the Kubernetes part.

invenio/values.yaml
```diff
@@ -72,6 +73,16 @@ web:
     min_web_replicas: 2
   assets:
     location: /opt/invenio/var/instance/static
+  # sometimes its desirable to set resources values here
```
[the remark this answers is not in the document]
wgresshoff: Didn't want to change your code too heavily. But I'm sure not anyone wants to define ephemeral-storage (it's checked by our new K8s cluster), so I comment it out, ok?!.

invenio/templates/pvc.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.persistence.enabled }}
+{{- if and .Values.persistence.enabled (not .Values.persistence.existing_pvc) }}
```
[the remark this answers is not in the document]
wgresshoff: You're right, that change isn't needed at all. Reverted.

invenio/templates/deployments/web.yaml
```diff
@@ -90,6 +90,9 @@ spec:
               key: {{ default (printf "INVENIO_%s_APP_CREDENTIALS" .name) }}
         {{- end }}
         {{- end }}
+        {{- if .Values.extraEnv }}
```
ppanero, line 93: I am wondering if instead of putting a *catch all* config here, we could put it in the `configurations/invenio.yaml`. Unless we are talking about secrets? Note that even if not an `INVENIO_` they would go to the config file (we do that for e.g. `FLASK_` variables.

invenio/templates/deployments/worker.yaml
```diff
@@ -80,6 +80,9 @@ spec:
               name: {{ .Values.invenio.datacite.secret_name }}
               key: DATACITE_PASSWORD
         {{- end }}
+        {{- if .Values.extraEnv }}
```
ppanero, line 83: same as above.

ppanero (commented):
Im not a fan of the `extraEnv` I think we are duplicating config loading... But maybe I am missing some use case.

I also have some very minor changes locally, otherwise we can merge once we agree on the `extraEnv`.

invenio/templates/deployments/worker.yaml
```diff
@@ -217,7 +223,8 @@ spec:
         {{- end }}
       volumes:
       - name: var-run-celery
-        emptyDir: {}
+        emptyDir:
+          medium: Memory
```
ppanero, line 227: Is this needed in the end? I seem to recall you said it did not fix the problem?

invenio/templates/deployments/web.yaml
wgresshoff: I thought of that solution, but in the end I opted for a way to do all configuration in just one place (the values.yaml). Just don't like the idea I have to change files from the chart itself and maintain all my changes through all next updates.

invenio/templates/deployments/worker.yaml
wgresshoff: Yes, you are right, but that's just half the truth. In the end with this definition we have the additional possibility of just deleting the pod. My expectation was that would work for automatic restart by Kubernetes, too. But perhaps there is a better solution we don't know yet. I would like to have that in if it does no harm because it helps operating the system.

ppanero: Sounds fair, thanks!

invenio/templates/deployments/web.yaml
ppanero: Sorry, I think I did not explain myself 😅. I do agree with config coming all the way from one place (values.yaml). My point was about *where to do the expansion*. Right now it is being don in `web.yaml`, `worker.yaml`, etc. which leads to repeated code and if needed to be changed without helm: editing a yaml file manually in the deployement. What I was suggesting was that we would have only:

`config.yaml`
```
        {{- if .Values.extraEnv }}
         {{- toYaml .Values.extraEnv | nindent 8 }}
         {{- end }}
```

WDYT?

wgresshoff: Great idea :+1: That makes it just more easy to maintain the chart! I'm in complete agreement!

ppanero:
@wgresshoff I merged the celery dir fix separately, there is only missing the extraEnv stuff... I will try to fix it while deploying tomorrow/friday

ppanero:
@wgresshoff unfortunately I did not have time to implement the changes

wgresshoff:
This is completely outdated and unneccessary now. So I'm closing this.

PR #49, opened 2022-10-28
nginx: fix missing value

jrcastro2:
Nginx conf was missing this value [here](https://github.com/inveniosoftware/helm-invenio/blob/master/invenio/templates/configurations/nginx.yaml#L65)


ppanero (commented):
The max body size was refactored and splitted between the records and files endpoint (the second might need to be larger). See https://github.com/inveniosoftware/helm-invenio/commit/8b4704de307673da47a899ca248ff755fd4a588a

I am wondering if instead of adding another related config var, we should instead change the config to use the `records.client_max_body_size`. WDYT?

kpsherva (approved):

Issue #53, opened 2022-11-15
INVENIO_DATACITE_ENABLED is "false" instead of "False" leading to a bad python evaluation


avivace:
We are using `.Values.invenio.datacite.enabled` both as a flag (e.g. to decide when to inject the datacite credentials) and as the value to be injected as `INVENIO_DATACITE_ENABLED` in the [invenio config map](https://github.com/inveniosoftware/helm-invenio/blob/master/invenio/templates/configurations/invenio.yaml#L13) (which feeds the environment). 
However, YAML's boolean `false` gets evaluated to true in python, leading to a messed up configuration (e.g. try to deploy the helm charts with the default values, datacite comes up as enabled because `INVENIO_DATACITE_ENABLED` is set as `false` and not `False` in the configMap).

avivace:
Currently waiting for some feedback - should this be escalated to invenio-config and how we parse invenio.cfg?

kpsherva:
could it be because of the quote marks around the value? it gets evaluated as string and then as true. Did you test without the quotes?

PR #54, opened 2022-11-15
Explicitly set boolean values with their python literals

avivace:
closes https://github.com/inveniosoftware/helm-invenio/issues/53

This is just a workaround - since invenio is evaluating the environment variables as Python literals, those should NOT used as YAML guard variables in the helm charts, or every "false" value will be evaluated to true leading to misconfigurations

### DIFF
--- a/invenio/templates/configurations/invenio.yaml
+++ b/invenio/templates/configurations/invenio.yaml
@@ -12,10 +12,14 @@ data:
   INVENIO_COLLECT_STORAGE: flask_collect.storage.file
   {{ if .Values.invenio.datacite.enabled }}
   INVENIO_DATACITE_PREFIX: "{{ .Values.invenio.datacite.prefix }}"
-  INVENIO_DATACITE_TEST_MODE: "{{ .Values.invenio.datacite.test_mode }}"
   INVENIO_DATACITE_ENABLED: "True"
   {{ else }}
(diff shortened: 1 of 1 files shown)


kpsherva (approved):

invenio/templates/configurations/invenio.yaml
```diff
@@ -10,10 +10,16 @@ data:
   INVENIO_CACHE_REDIS_URL: 'redis://{{ include "redis.host_name" . }}:6379/0'
   INVENIO_CELERY_RESULT_BACKEND: 'redis://{{ include "redis.host_name" . }}:6379/2'
   INVENIO_COLLECT_STORAGE: flask_collect.storage.file
-  INVENIO_DATACITE_ENABLED: "{{ .Values.invenio.datacite.enabled }}"
   {{ if .Values.invenio.datacite.enabled }}
   INVENIO_DATACITE_PREFIX: "{{ .Values.invenio.datacite.prefix }}"
-  INVENIO_DATACITE_TEST_MODE: "{{ .Values.invenio.datacite.test_mode }}"
+  INVENIO_DATACITE_ENABLED: "True"
```
zzacharo, line 15: minor: is there a way to have a small utility function/macro to return the correct boolean?

zzacharo (approved):

PR #56, opened 2022-12-01
worker: make celery beat only run scheduler

slint:
* The previous parameters spawned an actual worker which also run  beat/scheduler. Current call only spawns the `celery beat` service.

See also https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#starting-the-scheduler

### Description

Please describe briefly your pull request.

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**


invenio/templates/deployments/worker.yaml
```diff
@@ -145,7 +145,7 @@ spec:
         command: [
           "/bin/bash",
           "-c",
-          "celery -A {{ .Values.worker.app }} worker --beat -l {{ .Values.worker.log_level }}  -s {{ .Values.worker.celery_schedule }} --pidfile {{ .Values.worker.celery_pidfile }}"
+          "celery -A {{ .Values.worker.app }} beat -l {{ .Values.worker.log_level }} -s {{ .Values.worker.celery_schedule }} --pidfile {{ .Values.worker.celery_pidfile }}"
```
ntarocco, line 148: I am totally ok with this, we just kept the worker in other prod instances to avoid to "waste" a POD only for the beat and also have an extra worker, but making sure that the `replica` count will be always only `1`.

ppanero:
Would this mean that it would consume less resources? right now, with less than 2GB of memory it would crash... If we can run it with 1GB we would have one less worker (true) but in terms of architecture it would be clearer where the components doing a task are.

ntarocco:
> Would this mean that it would consume less resources? right now, with less than 2GB of memory it would crash... If we can run it with 1GB we would have one less worker (true) but in terms of architecture it would be clearer where the components doing a task are.

Just tested, worker with beat:

Screenshot 2022-12-13 at 16 16 17

Only beat:

Screenshot 2022-12-13 at 16 19 16

We can greatly decrease resources if we go with this. At the moment, the [defaults](https://github.com/inveniosoftware/helm-invenio/blob/master/invenio/values.yaml#L126) are still very high.

ppanero:
Thanks a lot for testing that!!!! While I do think we would be "wasting a pod", since we have more than enough to do a rollout, if we can save resources... Do you see any benefits of having worker + beat?

ntarocco:
I am totally ok to with a POD with only `beat`, it makes a lot of sense.
@slint if you can simply maybe update the resources for the `workerBeat` values, then it is perfect... we could try to request 200Mb of mem, limit 500?

slint:
Done, memory limits for `worker-beat` updated to 200/500 Mb

slint:
Added, sorry had no idea this existed :)

PR #59, opened 2022-12-14
configurations: support invenio config injection

ppanero:
- closes https://github.com/zenodo/zenodo-rdm/issues/113
- example: https://gitlab.cern.ch/zenodo/zenodo-rdm-openshift/-/merge_requests/1

Helm Charts do not accept injecting external files, as a security feature. We could still have the values in different files, but given that the way of injecting them would be the same we might as well just have one `values.yaml` file.

### DIFF
--- a/invenio/templates/configurations/invenio.yaml
+++ b/invenio/templates/configurations/invenio.yaml
@@ -9,31 +9,11 @@ data:
   INVENIO_CACHE_REDIS_HOST: '{{ include "redis.host_name" . }}'
   INVENIO_CACHE_REDIS_URL: 'redis://{{ include "redis.host_name" . }}:6379/0'
   INVENIO_CELERY_RESULT_BACKEND: 'redis://{{ include "redis.host_name" . }}:6379/2'
-  INVENIO_COLLECT_STORAGE: flask_collect.storage.file
-  {{ if .Values.invenio.datacite.enabled }}
-  INVENIO_DATACITE_PREFIX: "{{ .Values.invenio.datacite.prefix }}"
-  INVENIO_DATACITE_ENABLED: "True"
-  {{ else }}
-  INVENIO_DATACITE_ENABLED: "False"
-  {{ end }}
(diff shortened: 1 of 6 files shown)


invenio/templates/configurations/invenio.yaml
```diff
@@ -9,31 +9,11 @@ data:
...
-  {{ if .Values.search.enabled }}
   INVENIO_SEARCH_HOSTS: "{{ .Values.search.invenio_hosts }}"
-  {{ end }}
-  INVENIO_SEARCH_INDEX_PREFIX: "{{ .Values.invenio.search.index_prefix }}"
   INVENIO_SITE_HOSTNAME: '{{ .Values.host }}'
   INVENIO_SITE_UI_URL: 'https://{{ .Values.host }}'
   INVENIO_SITE_API_URL: 'https://{{ .Values.host }}/api'
-  SENTRY_ENVIRONMENT: "{{ .Values.invenio.logging.sentry.environment }}"
+  {{- range $key, $value := .Values.invenio.extra_config }}
+  {{ $key }}: {{ $value }}
```
ppanero, line 18: This implementation has an issue with complex k,v pairs (e.g. if wanting to overwrite CSP)

invenio/values.yaml
```diff
@@ -49,6 +32,11 @@ invenio:
       - name: ""
         consumer_key: ""
         consumer_secret: ""
+  extra_config:
+    INVENIO_COLLECT_STORAGE: "flask_collect.storage.file"
+    INVENIO_DATACITE_ENABLED: "False"
```
ppanero, line 37: The only hassle is that due to how values are injected in the config, it expects a string so numbers and booleans need to be wrapped around with quotes. The result configMap is correct, i.e. the app understands correctly bools and numbers. (tested in zenodo-rdm-qa).

ppanero:
When merging, create an issue to update the values on demoinvenio-rdm-openshift (gitlab)

ntarocco (approved):

PR #65, opened 2023-05-09
PVC: add redis and rabbitMQ PVC

jrcastro2:


kpsherva:
silly question: could you rename the commit to
`deployments: add redis and rabbitmq persistent volume claims` 
? I am sure it will be easier to search in the history than the acronym :)

kpsherva (approved):

invenio/values.yaml
```diff
@@ -127,6 +127,20 @@ persistence:
   size: 10G
   storage_class: ""
 
+mq_persistence:
```
ntarocco, line 130: I would propose to move these and the redis one inside related keys, e.g.:
```
redis:
    ...
    persistency:
        enabled: 
        ...

File: invenio/values.yaml [hunk 2]
```diff
@@ -127,6 +127,20 @@ persistence:
...
+  enabled: true
+  name: "rabbitmq-volume"
+  access_mode: ReadWriteMany
+  size: 10G
+  storage_class: "cephfs-ssd-no-backup"
```
ntarocco, line 135: This is a CERN value. As discussed, maybe make it `""`? (it should take the default)
We need then to override this in our CERN configs and use this value there.

invenio/values.yaml
```diff
@@ -139,6 +146,12 @@ rabbitmq:
   secret_name: "mq-secrets"
   default_password: "mq_password"
   celery_broker_uri: "amqp://guest:mq_password@mq:5672/"
+  persistence:
+    enabled: true
+    name: "rabbitmq-volume"
+    access_mode: ReadWriteMany
+    size: 10G
+    storage_class: ""
```
kpsherva, line 154: whats the consequence of the empty string as a storage_class? is it an optional param?

invenio/values.yaml
```diff
@@ -131,6 +131,13 @@ redis:
   enabled: true
   image: "redis:7"
   host: ""
+  persistence:
+    enabled: true
+    name: "redis-volume"
+    access_mode: ReadWriteMany
```
ntarocco, line 137: Minor: it is probably enough `ReadWriteOnce`, given that it is not shared. Anyway, we can keep it like this.

ntarocco (approved):

Issue #68, opened 2023-05-31
Missing rewrite rule for dots in the end of the domain name


avivace:
Trying to access a deployment of inveniordm (such as [zenodo-rdm](https://zenodo-rdm.web.cern.ch./) or the [sandbox](https://inveniordm.web.cern.ch./) adding a "dot" at the end of the domain name returns a server error (because the host is not trusted) instead of redirecting to the base domain.

Probably a missing RewriteRule in the [nginx configuration](https://github.com/inveniosoftware/helm-invenio/blob/master/invenio/templates/configurations/nginx.yaml)?

PR #69, opened 2023-05-31
nginx: fix trailing dots in the url

kpsherva:
* closes https://github.com/inveniosoftware/helm-invenio/issues/68

### DIFF
--- a/invenio/templates/configurations/nginx.yaml
+++ b/invenio/templates/configurations/nginx.yaml
@@ -39,6 +39,10 @@ data:
         # in case of chunked transfer encoding
         proxy_http_version 1.1;
 
+        if ($http_host ~ "\.$" ){
+          rewrite ^(.*) $scheme://$host$1 permanent;
+        }
+
         location /ping {
           access_log off;
           return 200 "ok\n";


avivace (approved):

ntarocco:
I don't see much the point in doing this, given that any other char could be added at the end in a sentence (`,`,`;` etc...)

avivace:
> I don't see much the point in doing this, given that any other char could be added at the end in a sentence (`,`,`;` etc...) If you really want to do it, then it is for InvenioRDM demo instance only, not generic helm charts: you cannot know in advance if this is wanted or not.

No, adding any other character simply won't resolve the domain (your browser will just give up or throw a search query). A trailing dot, though, is still valid (actually, fully-qualified DNS domain names [do have a dot at the end](http://www.dns-sd.org/trailingdotsindomainnames.html)) and will get resolved by the DNS record; that's why we were getting InvenioRDM server errors.

ntarocco:
> > I don't see much the point in doing this, given that any other char could be added at the end in a sentence (`,`,`;` etc...) If you really want to do it, then it is for InvenioRDM demo instance only, not generic helm charts: you cannot know in advance if this is wanted or not.
> 
> Adding any other character simply won't resolve the domain (your browser will just give up or throw a search query). A trailing dot, though, is still valid (actually, fully-qualified DNS domain names [do have a dot at the end](http://www.dns-sd.org/trailingdotsindomainnames.html)) and will get resolved by the DNS record; that's why we were getting InvenioRDM server errors.

Sorry, I was not clear with what I wrote. Let me try again.

Here it is not about fully-qualified domain. This change was triggered by a wrong link in a blog post: by mistake, a `.` was added at the end. If a `,` was added instead, we would not discuss this.
We cannot blindly add a change like this one, because:
- if tested in our env (at CERN) and it works, we cannot assume it will work anywhere else
- we don't want to force any user to have this by default, and with this config. What is someone else does not want to handle this in all the `nginx` containers, but closer to the LBs instead? There might be also issue with the backup website in the future.
- we are not assessing the consequences/caveats enough, see for example [here](https://webmasters.stackexchange.com/a/74099) and [here](https://serverfault.com/questions/470719/https-and-trailing-dot-in-domain).

IMO, this introduces more risks than solving issues: on top of that, it might solve the 0,0***01% of the wrong links.

If you really want to propose a solution to the `.` in the domain:
- Write this in a recipe in the doc. We have a dedicated deployment section in the doc.
- Apply this to your infrastructures only.
- Make this optional, disabled by default, in the generic helm charts. I would personally not do that. We already more than enough configurability, and we will add more.

TL;DR: In the generic context of the Invenio Helm charts, this feels a rushed workaround, which might introduce more problems than solving them.

kpsherva:
> > > I don't see much the point in doing this, given that any other char could be added at the end in a sentence (`,`,`;` etc...) If you really want to do it, then it is for InvenioRDM demo instance only, not generic helm charts: you cannot know in advance if this is wanted or not.
> > 
> > 
> > Adding any other character simply won't resolve the domain (your browser will just give up or throw a search query). A trailing dot, though, is still valid (actually, fully-qualified DNS domain names [do have a dot at the end](http://www.dns-sd.org/trailingdotsindomainnames.html)) and will get resolved by the DNS record; that's why we were getting InvenioRDM server errors.
> 
> Sorry, I was not clear with what I wrote. Let me try again.
> 
> Here it is not about fully-qualified domain. This change was triggered by a wrong link in a blog post: by mistake, a `.` was added at the end. If a `,` was added instead, we would not discuss this. We cannot blindly add a change like this one, because:
> 
> * if tested in our env (at CERN) and it works, we cannot assume it will work anywhere else
> * we don't want to force any user to have this by default, and with this config. What is someone else does not want to handle this in all the `nginx` containers, but closer to the LBs instead? There might be also issue with the backup website in the future.
> * we are not assessing the consequences/caveats enough, see for example [here](https://webmasters.stackexchange.com/a/74099) and [here](https://serverfault.com/questions/470719/https-and-trailing-dot-in-domain).
> 
> IMO, this introduces more risks than solving issues: on top of that, it might solve the 0,0***01% of the wrong links.
> 
> If you really want to propose a solution to the `.` in the domain:
> 
> * Write this in a recipe in the doc. We have a dedicated deployment section in the doc.
> * Apply this to your infrastructures only.
> * Make this optional, disabled by default, in the generic helm charts. I would personally not do that. We already more than enough configurability, and we will add more.
> 
> TL;DR: In the generic context of the Invenio Helm charts, this feels a rushed workaround, which might introduce more problems than solving them.

I agree with @ntarocco , I would add it to the deployment docs instead, since it is corner case. I've applied it only on zenodo-rdm for the time being. I would add it as configurable because I agree that we already have quite overwhelming amount of configuration points.

PR #76, opened 2023-09-22
Improve config

ntarocco:


invenio/templates/configurations/uwsgi.yaml
```diff
@@ -6,31 +6,47 @@ metadata:
...
+    single-interpreter = true
+    need-app = true
+    lazy = true
     lazy-apps = true
-    # thunder-lock = true # https://marc.info/?l=uwsgi&m=140473636200986&w=2
-                          # https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html
+
+    # Silence write errors for misbehaving clients
+    # https://github.com/getsentry/raven-python/issues/732
+    # ignore-sigpipe = true
```
alejandromumo, line 33: Just to be sure, these were added due to `IOError` but are commented. Is that ok?

alejandromumo (approved):
Thanks a lot! Left a minor comment

invenio/templates/configurations/uwsgi.yaml
```diff
@@ -6,31 +6,47 @@ metadata:
...
+    # ignore-write-errors = true
+    # disable-write-exception = true
+
+    # post-buffering = true
+    buffer-size = 65535
```
kpsherva, line 38: not a blocker, just for my personal knowledge
how were those values calculated? are they standard recommended ?

invenio/templates/configurations/uwsgi.yaml
ntarocco: This is a copy/paste from what has been set in ZenodoRDM OpenShift.
Any settings should be/will be re-checked.

invenio/templates/configurations/uwsgi.yaml
ntarocco: See comment above

invenio/templates/deployments/web.yaml
```diff
@@ -111,28 +111,17 @@ spec:
               name: {{ .valueFrom.secretKeyRef.name }}
               key: {{ .valueFrom.secretKeyRef.key }}
         {{- end }}
-        readinessProbe:
```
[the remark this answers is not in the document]
ntarocco: No, we want :) We can chat IRL

PR #82, opened 2023-10-12
haproxy: add config for extra http_request statements

zzacharo:


invenio/templates/configurations/haproxy.yaml
```diff
@@ -46,6 +46,8 @@ data:
       http-request deny if is_blocked_ip OR is_blocked_ua
       errorfiles custom_errors
 
+      {{ .Values.haproxy.extra_frontend_public_http_request| indent 6 }}
```
ntarocco, line 49: if you put `indent 6`, I think you have to put this all to the left

alejandromumo (approved):

invenio/templates/configurations/haproxy.yaml
alejandromumo: Tested this and the changes are also OK. We're merging for now since it's working.

PR #92, opened 2023-12-04
Parameterize timezone variable

lindhe:
### Description

This change parameterizes the timezone variable used across deployments. The new value is `global.timezone`, following [Bitnamis pattern](https://github.com/bitnami/charts/tree/main/bitnami/apache) of having a `globals` object for values that are used across multiple components.

### Checklist

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

### DIFF
--- a/charts/invenio/templates/install-init-job.yaml
+++ b/charts/invenio/templates/install-init-job.yaml
@@ -34,7 +34,7 @@ spec:
             name: invenio-config
         env:
         - name: TZ
-          value: {{ .Values.global.timezone }}
+          value: {{ required "Missing .Values.global.timezone" .Values.global.timezone }}
         - name: INVENIO_BROKER_URL
           valueFrom:
             secretKeyRef:
--- a/charts/invenio/templates/install-init-job.yaml
+++ b/charts/invenio/templates/install-init-job.yaml
@@ -34,7 +34,7 @@ spec:
             name: invenio-config
         env:
         - name: TZ
-          value: "Europe/Zurich"
+          value: {{ .Values.global.timezone }}
         - name: INVENIO_BROKER_URL
           valueFrom:
             secretKeyRef:
--- a/charts/invenio/templates/flower/deployment.yaml
+++ b/charts/invenio/templates/flower/deployment.yaml
@@ -31,7 +31,7 @@ spec:
             - containerPort: 5555
           env:
             - name: TZ
-              value: {{ .Values.global.timezone }}
+              value: {{ required "Missing .Values.global.timezone" .Values.global.timezone }}
             - name: RABBITMQ_DEFAULT_PASS
               valueFrom:
                 secretKeyRef:
--- a/charts/invenio/templates/flower/deployment.yaml
+++ b/charts/invenio/templates/flower/deployment.yaml
@@ -31,7 +31,7 @@ spec:
             - containerPort: 5555
           env:
             - name: TZ
-              value: "Europe/Zurich"
+              value: {{ .Values.global.timezone }}
             - name: RABBITMQ_DEFAULT_PASS
               valueFrom:
                 secretKeyRef:
--- a/charts/invenio/templates/opensearch/deployment.yaml
+++ b/charts/invenio/templates/opensearch/deployment.yaml
@@ -31,7 +31,7 @@ spec:
               memory: 1Gi
           env:
             - name: TZ
-              value: {{ .Values.global.timezone }}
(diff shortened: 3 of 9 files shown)


plesoun-stein:
it's good idea, I like it and have an improvement. We need to set some default, 
```
value: {{ .Values.invenio.timezone | default "Europe/Zurich" | quote }}
```
With invenio as subchart we need to set a global.timezone in main chart. It's prone to errors. 

invenio.timezone may be better location as it may be referenced in main chart or subchart. 

And with a default value you can omit setting without error.

lindhe:
> […] It's prone to errors.
> […]
> And with a default value you can omit setting without error.

I agree, we should 100% handle the case when the timezone is unset! Very sloppy of me to not handle that in my original PR, sorry.

I usually prefer using `required` rather than `default`. That forces the user to be explicit with their values. It is easier to document; just look in the `values.yaml` to see what the default value is, rather than 10 different places, which works well with tools like [Readme Generator For Helm](https://github.com/bitnami/readme-generator-for-helm). Also, I find that using the `default` function introduces implicit values which may be harder to understand and debug for the end user than the values they are forced to set explicitly. Do you agree?

> invenio.timezone may be better location as it may be referenced in main chart or subchart.

Unless I misunderstand what you mean, it is absolutely possible to set `global.timezone` even when Invenio is included as a subchart! Like this, if I recall correctly:

**Chart.yaml**
```yaml
apiVersion: v2
name: foo
description: An umbrella chart that includes subcharts.
dependencies:
  - name: invenio
    version: "1.2.3"
    repository: "https://example.com/charts"
```

**values.yaml**
```yaml
invenio:
  global:
    timezone: Europe/Stockholm
```

plesoun-stein:
yes, required is better. definitely. 

I'm sorry, i wasn;t accurate with `global`

You have `.Values.global.timezone` in your commit. It is use of helm builtin `.Values.global`

 `.Values.invenio.global.timezone` in your example, It is subkey of invenio definitions. 

but both are good option

lindhe:
I've made the value required now. Do you want me to also drop the `global` object or not?

**A:**

```yaml
# values.yaml
global:
  timezone: "Europe/Zurich"
```

**B:**
```yaml
# values.yaml
timezone: "Europe/Zurich"
```

**Do you want (A) or (B)?**

charts/invenio/templates/install-init-job.yaml
```diff
@@ -34,7 +34,7 @@ spec:
             name: invenio-config
         env:
         - name: TZ
-          value: "Europe/Zurich"
+          value: {{ required "Missing required "Missing .Values.global.timezone" .Values.global.timezone" .Values.global.timezone }}
```
lindhe, line 37: Wow, what's happening here? Looks like a typo, sorry. I'll fix.

lindhe: Same issue with a few other places. Should have proof read my commit first…

plesoun-stein:
global object should be better for chart with subcharts. 

thanks.

plesoun-stein:
A variant.

plesoun-stein:
yes, "A" variant

lindhe:
Ahaaa, right! Thanks, I get it now. Thought you meant "a variant" and I did not get it at all. 😄

I think it's already the (A) variant. So all good?

plesoun-stein:
yes. :-)

PR #125, opened 2024-08-16
Support for custom AMQP protocol and external INVENIO_SEARCH_HOSTS

fradeve:
### Description

This PR solves 2 issues with the Helm chart

#### RabbitMQ protocol can't be customized

The default protocol for RabbitMQ is `amqp` and can't be edited -- this is probably fine when running RabbitMQ locally, but most _hosted_ RMQ instances will only accept connections on `amqps`; this PR adds support for a custom protocol when using `.Values.rabbitmqExternal`

#### Impossible to define custom Search username, password or port

The easiest way to use an external {Open/Elastic}Search instance is to pass `INVENIO_SEARCH_HOSTS` in the environment; however, the way `INVENIO_SEARCH_HOSTS` is set in the Configmap:

```
INVENIO_SEARCH_HOSTS: {{ printf "[{'host': '%s'}]" (include "invenio.opensearch.hostname" .) | quote }}
```

 means that only the `hostname` can be configured, without port, user or password; this is of course unworkable when using an hosted {Elastic/Open}Search that needs proper authentication to be passed to the application.

A possible way around this is to add the `INVENIO_SEARCH_HOSTS` string into `Values.invenio.extra_config`; this will however force Helm to have 2 `INVENIO_SEARCH_HOSTS` in the Configmap -- which won't work either.

In this PR, if `INVENIO_SEARCH_HOSTS` is set in `extra_config`, it is removed from the Configmap so we don't get this defined twice, avoiding clashes and allowing the app to connect to an externally hosted search engine.

### Checklist

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] ~~I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).~~
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Frontend**

- [ ] ~~I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.~~
- [ ] ~~I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.~~
- [ ] ~~I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.~~

**Reminder**

3. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -108,9 +108,7 @@
   This template renders the protocol for RabbitMQ.
 */}}
 {{- define "invenio.rabbitmq.protocol" -}}
-  {{- if .Values.rabbitmq.enabled }}
-    {{- include "common.names.fullname" .Subcharts.rabbitmq -}}
-  {{- else }}
+  {{- if .Values.rabbitmq.enabled }}amqp{{- else }}
     {{- required "Missing .Values.rabbitmqExternal.protocol" .Values.rabbitmqExternal.protocol }}
   {{- end }}
 {{- end -}}
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -103,6 +103,18 @@
   {{- end }}
 {{- end -}}
 
+##########################     RabbitMQ protocol     ##########################
+{{/*
+  This template renders the protocol for RabbitMQ.
+*/}}
+{{- define "invenio.rabbitmq.protocol" -}}
+  {{- if .Values.rabbitmq.enabled }}
+    {{- include "common.names.fullname" .Subcharts.rabbitmq -}}
+  {{- else }}
+    {{- required "Missing .Values.rabbitmqExternal.protocol" .Values.rabbitmqExternal.protocol }}
+  {{- end }}
+{{- end -}}
+
 ##########################     Celery broker URI     ##########################
 {{/*
   This template renders the URI for connecting to RabbitMQ.
@@ -112,7 +124,8 @@
   {{- $password := (include "invenio.rabbitmq.password" .) -}}
   {{- $port := (include "invenio.rabbitmq.amqpPort" .) -}}
   {{- $hostname := (include "invenio.rabbitmq.hostname" .) -}}
-  {{- printf "amqp://%s:%s@%s:%v/" $username $password $hostname $port }}
+  {{- $protocol := (include "invenio.rabbitmq.protocol" .) -}}
+  {{- printf "%s://%s:%s@%s:%v/" $protocol $username $password $hostname $port }}
 {{- end -}}
 
 ###########################     RabbitMQ API URI     ###########################
--- a/charts/invenio/templates/invenio-configmap.yaml
+++ b/charts/invenio/templates/invenio-configmap.yaml
@@ -11,7 +11,10 @@ data:
   INVENIO_CELERY_RESULT_BACKEND: 'redis://{{ include "invenio.redis.hostname" . }}:6379/2'
   INVENIO_RATELIMIT_STORAGE_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/3'
   INVENIO_COMMUNITIES_IDENTITIES_CACHE_REDIS_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/4'
+  {{- if hasKey .Values.invenio.extra_config "INVENIO_SEARCH_HOSTS" }}
+  {{- else }}
   INVENIO_SEARCH_HOSTS: {{ printf "[{'host': '%s'}]" (include "invenio.opensearch.hostname" .) | quote }}
+  {{- end }}
   INVENIO_SITE_HOSTNAME: '{{ include "invenio.hostname" $ }}'
   INVENIO_SITE_UI_URL: 'https://{{ include "invenio.hostname" $ }}'
   INVENIO_SITE_API_URL: 'https://{{ include "invenio.hostname" $ }}/api'


charts/invenio/values.yaml
```diff
@@ -124,6 +124,8 @@ web:
       cpu: 1000m
       memory: 1Gi
   annotations: []
+  service:
+    type: NodePort
```
lindhe, line 128: This would be a breaking change to the chart's default behavior. I think it would be better to let the default value be empty (`type: ""`). You have already defined `ClusterIP` as default using the `default` function in the template, so having another default value here is a bit confusing.

charts/invenio/templates/invenio-configmap.yaml
```diff
@@ -11,7 +11,10 @@ data:
   INVENIO_CELERY_RESULT_BACKEND: 'redis://{{ include "invenio.redis.hostname" . }}:6379/2'
   INVENIO_RATELIMIT_STORAGE_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/3'
   INVENIO_COMMUNITIES_IDENTITIES_CACHE_REDIS_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/4'
+  {{- if hasKey .Values.invenio.extra_config "INVENIO_SEARCH_HOSTS" }}
+  {{- else }}
```
lindhe, line 15: Instead of having an empty if statement, I think it is preferable to negate the condition using [`not`](https://helm.sh/docs/chart_template_guide/function_list/#not).

charts/invenio/templates/_helpers.tpl
```diff
@@ -103,6 +103,16 @@
   {{- end }}
 {{- end -}}
 
+##########################     RabbitMQ protocol     ##########################
+{{/*
+  This template renders the protocol for RabbitMQ.
+*/}}
+{{- define "invenio.rabbitmq.protocol" -}}
+  {{- if .Values.rabbitmq.enabled }}amqp{{- else }}
```
lindhe, line 111: Please use the same style as elsewhere in the code. Example: https://github.com/inveniosoftware/helm-invenio/blob/2e2b043963edb0e37c507996419c195ddb21f1c1/charts/invenio/templates/_helpers.tpl#L70-L80

Samk13:
Thank you for your contribution. However, there's a [PR](https://github.com/inveniosoftware/invenio-app-rdm/pull/2918) introducing changes where the application constructs connection strings from environment variables. 
If merged, this might make the modifications in this PR redundant or require adjustments to align with the new approach. @lindhe could you please take a look at this?

lindhe:
You are right that our planned changes to connection strings would change things relating to this PR. But I don't think it's worth waiting for that.

If we merge this first, it's just a little more work for our future improvement. But if we wait for that first, than it risks delaying these improvements unnecessarily.

So I think we should not block this PR, but it's worth a heads-up that related changes are coming. And that may or may not be breaking changes in the Helm chart, depending on how easy it is for me to be backwards compatible with the change.

lindhe (changes requested):
I think this is looking very promising. I don't see the issue, but when I run `helm template foo ./charts/invenio -f ~/tmp/invenio-foo-values.yaml` it crashes. Here are the values I use:

```yaml
invenio:
  hostname: foo
  secret_key: foo
  security_login_salt: foo
  csrf_secret_salt: foo
postgresql:
  auth:
    password: foo
rabbitmq:
  enabled: true
  auth:
    password: foo
```


I'll try to debug later, but I don't have time right now. Please see if you can find the bug before I do. 🙃

charts/invenio/templates/web-service.yaml
```diff
@@ -15,3 +15,4 @@ spec:
       targetPort: 8080
   selector:
     app: web
+  type: {{ .Values.web.service.type }}
```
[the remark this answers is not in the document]
egabancho: I added `ClusterIP` as the default value ⬇️. Do you think we still need the required?

charts/invenio/templates/_helpers.tpl
```diff
@@ -103,6 +103,30 @@
   {{- end }}
 {{- end -}}
 
+##########################     RabbitMQ protocol     ##########################
+{{/*
+  This template renders the protocol for RabbitMQ.
+*/}}
+{{- define "invenio.rabbitmq.protocol" -}}
+  {{- if .Values.rabbitmq.enabled }}
+    "amqp"
```
lindhe, line 112:
```suggestion
    {{- "amqp" }}
```

charts/invenio/templates/_helpers.tpl
```diff
@@ -103,6 +103,30 @@
...
+  {{- end }}
+{{- end -}}
+
+##########################     RabbitMQ vhost     ##########################
+{{/*
+  This template renders the vhost for RabbitMQ.
+*/}}
+{{- define "invenio.rabbitmq.vhost" -}}
+  {{- if .Values.rabbitmq.enabled }}
+    ""
```
lindhe, line 124:
```suggestion
    {{- "" }}
```

charts/invenio/templates/web-service.yaml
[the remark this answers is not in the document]
lindhe: True. Then this is less important. If someone sets it to null, we can probably let that be their problem.

lindhe:
I finally got around to testing my changes. Having bare strings will break things if `rabbitmq.enabled=true`. But if we wrap them in template strings according to my suggestions, it works. If you fix that, I'll approve. :)

lindhe (approved):
I think it looks good now, thank you for your contribution!

Next time it would be great if you created one PR per logical change. Despite you doing exemplary work by creating three logically separate commits, GitHub's review system really only lets us review the aggregate changes. Also there's only a single PR title that people can search for, while your PR has three changes. So one PR per logical change would be even more helpful.

But that's for next time. This time: well done! 👍

PR #129, opened 2024-10-29
invenio: Add option to enable/disable podSecurityContext

mb-wali:

### Description

In OpenShift, where the user ID (UID) and group ID (GID) are automatically managed, the ability to remove these settings from the pod security context simplifies deployment

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/charts/invenio/templates/web-deployment.yaml
+++ b/charts/invenio/templates/web-deployment.yaml
@@ -214,8 +214,8 @@ spec:
               cpu: '1'
               memory: 100Mi
             requests:
-              cpu: 500m
-              memory: 50Mi
+              cpu: '1'
+              memory: 100Mi
           securityContext:
             allowPrivilegeEscalation: false
             capabilities:
--- a/charts/invenio/templates/web-deployment.yaml
+++ b/charts/invenio/templates/web-deployment.yaml
@@ -214,8 +214,8 @@ spec:
               cpu: '1'
               memory: 100Mi
             requests:
-              cpu: '1'
-              memory: 100Mi
+              cpu: 500m
+              memory: 50Mi
           securityContext:
             allowPrivilegeEscalation: false
             capabilities:
@@ -237,12 +237,14 @@ spec:
             - name: kerberos-credentials-cache
               mountPath: /tmp
         {{- end }}
+      {{- if .Values.invenio.podSecurityContext.enabled }}
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
         runAsGroup: 1000
         seccompProfile:
           type: "RuntimeDefault"
+      {{- end }}
       volumes:
       {{- if .Values.logstash.enabled }}
       - name: config
--- a/charts/invenio/templates/flower/deployment.yaml
+++ b/charts/invenio/templates/flower/deployment.yaml
@@ -50,12 +50,14 @@ spec:
               mountPath: /var/celery
             - name: flower-config-volume
               mountPath: /var/flower
+      {{- if .Values.invenio.podSecurityContext.enabled }}
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
         runAsGroup: 1000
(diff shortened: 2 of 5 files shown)


ntarocco:
@lindhe @Samk13 can you please have a look?

lindhe:
@ntarocco I can take it. Don't think I can add myself as reviewer or assignee, but you can consider me as such.

charts/invenio/templates/web-deployment.yaml
```diff
@@ -214,8 +214,8 @@ spec:
               cpu: '1'
               memory: 100Mi
             requests:
-              cpu: '1'
-              memory: 100Mi
+              cpu: 500m
+              memory: 50Mi
```
lindhe, line 218: These changes seem unrelated to the PR. Please revert them.

lindhe (changes requested):
I like the intention of this PR – giving the users the option to enable or disable the security context is good. And I think your changes look good; they do what you intend them to do (as far as I can tell). Just need to restore the resource requests that accidentally got included, according to my comment.

However, I think it would be even more useful if we gave the users the ability to fully configure the security contexts rather than just an on/off switch. Preferably configurable for each pod and container, similar to [how Bitnami does it](https://github.com/bitnami/charts/blob/0868c633cc6b2f16efcbbd86e0dce126ac0fdcd1/bitnami/postgresql/values.yaml#L481-L520). Is it clear to you how I mean?

Would you be willing to rewrite the values in such a manner instead? I think that would be more valuable to this project. If that's too much work, I think it can be retrofitted later when the need arises.

charts/invenio/templates/web-deployment.yaml
mb-wali: You're absolutely correct; these changes are unrelated to the PR. However, I noticed that the current resource requests seemed a bit high for an init container, so I made a small adjustment to optimize it. Later, it might be helpful to define this in the values, just like the parent container resources. Let me know if you'd still like me to revert these changes.

mb-wali:
> I like the intention of this PR – giving the users the option to enable or disable the security context is good. And I think your changes look good; they do what you intend them to do (as far as I can tell). Just need to restore the resource requests that accidentally got included, according to my comment.
> 
> However, I think it would be even more useful if we gave the users the ability to fully configure the security contexts rather than just an on/off switch. Preferably configurable for each pod and container, similar to [how Bitnami does it](https://github.com/bitnami/charts/blob/0868c633cc6b2f16efcbbd86e0dce126ac0fdcd1/bitnami/postgresql/values.yaml#L481-L520). Is it clear to you how I mean?
> 
> Would you be willing to rewrite the values in such a manner instead? I think that would be more valuable to this project. If that's too much work, I think it can be retrofitted later when the need arises.

Certainly, I can see the value in making the security context fully configurable, as you suggested, similar to Bitnami’s approach. I'd be happy to proceed with the current scope of the PR for now and then open an issue to address a more comprehensive configuration in the future. This way, we can implement the core changes immediately and expand on them later when the need arises. Thanks for the thoughtful feedback!

charts/invenio/templates/web-deployment.yaml
lindhe: I agree that they can most likely be improved. But that improvement belongs in a separate PR. 👍

PR #130, opened 2024-10-30
web: initContainers resources from values

mb-wali:

### Description

Adding possibility to change resources for initContainers from values.yaml

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**


Samk13:
LGTM, makes `initContainers` resources configurable, improving flexibility.
@lindhe  tagging you for review as our DevOps expert as I am unable to assign you on this one.

charts/invenio/templates/web-deployment.yaml
```diff
@@ -209,13 +209,10 @@ spec:
               "-c",
               "cp -R {{ .Values.web.assets.location }}/. /opt/nginx-invenio-assets/",
           ]
+         {{- if .Values.web.initContainers.resources }}
```
lindhe, line 212: This will fail if `.Values.web.initContainers=null`. To catch this, we can wrap the existing if-statement in `{{- if .Values.web.initContainers }}`.

It's not a terribly likely scenario, considering you have set a good default value. But I think it's good to try and catch potential failure modes before they happen. Helm charts can be a hard to debug otherwise.

On the other hand, we are not catching `.Values.web=null` either. I just wanted to comment on this since this is new behavior and we have the opportunity to get it right from the start.

You decide if we catch this error or merge as-is!

lindhe (approved):
Please review my comment. Except for that: LGTM.

charts/invenio/templates/web-deployment.yaml
mb-wali: Thank you for your comment! You raise a valid point regarding the potential for defining new properties for `initContainers`. If we only check for `.Values.web.initContainers`, it could inadvertently add resources if a user defines new properties without explicitly wanting to include any resource definitions.

For instance, if a user specifies an initContainer with just a name but does not wish to define resources, it could look something like this:
```yaml
initContainers:
  - name: my-init-container
    # Resources are not defined here
```    
I think it’s important to ensure that we don’t unintentionally enable resource definitions just by adding new properties to the `initContainers`. Instead, we could consider a more granular approach that allows users to define initContainers with specific properties, including the ability to control resource allocation independently.

I appreciate your perspective on this, and I’ll take it into account as we move forward. If we decide to implement checks for` .Values.web.initContainers,` we should ensure that it doesn’t override the user’s intent regarding resource definitions. Thank you for your thoughtful input!

Samk13:
Given The potential edge case is minimal, and given the good default values already in place, it’s unlikely to break anything significant. I will merge it now.
Thanks for this!

Samk13 (approved):

PR #132, opened 2024-11-29
Templated variables inside invenio-config configMap

egabancho:
The first commit is the one that contains the change. 

The second one is a "test" to see what you think about renaming all the non-camelcase variables we have and deprecating them.

### DIFF
--- a/charts/invenio/templates/NOTES.txt
+++ b/charts/invenio/templates/NOTES.txt
@@ -1,1 +1,11 @@
 Invenio is ready to rock 🚀
+
+Your release is named {{ .Release.Name }}.
+
+{{- if .Values.invenio.extra_config}}
+
+DEPRECATION WARNING: 
+    `invenio.extra_config` has been renamed to `invenio.extraConfig` and will be
+    removed in a future release.
+
+{{- end }}
--- a/charts/invenio/templates/invenio-configmap.yaml
+++ b/charts/invenio/templates/invenio-configmap.yaml
@@ -20,5 +20,5 @@ data:
   INVENIO_DATACITE_ENABLED: "False"
   INVENIO_LOGGING_CONSOLE_LEVEL: "WARNING"
   {{- range $key, $value := .Values.invenio.extra_config }}
-  {{ $key }}: {{ $value | toYaml | indent 4 | trim }}
+  {{ $key }}: {{ tpl $value $ | toYaml | indent 5 | trim }}
   {{- end }}
--- a/charts/invenio/templates/invenio-configmap.yaml
+++ b/charts/invenio/templates/invenio-configmap.yaml
@@ -20,5 +20,8 @@ data:
   INVENIO_DATACITE_ENABLED: "False"
   INVENIO_LOGGING_CONSOLE_LEVEL: "WARNING"
   {{- range $key, $value := .Values.invenio.extra_config }}
-  {{ $key }}: {{ tpl $value $ | toYaml | indent 5 | trim }}
+  {{ $key }}: {{ $value | toYaml | indent 4 | trim }}
+  {{- end }}
+  {{- range $key, $value := .Values.invenio.extraConfig }}
+  {{ $key }}: {{ tpl $value $ | toYaml | indent 4 | trim }}
   {{- end }}
--- a/charts/invenio/values.yaml
+++ b/charts/invenio/values.yaml
@@ -37,6 +37,8 @@ invenio:
       - name: ""
         consumer_key: ""
         consumer_secret: ""
+  ## @param invenio.extra_config Extra environment variables (templated) to be added to all the pods.
+  ##
   extra_config: {}
   extra_env_from_secret: []
     # - name: NAME_OF_MY_ENV_VAR
--- a/charts/invenio/values.yaml
+++ b/charts/invenio/values.yaml
@@ -37,9 +37,11 @@ invenio:
       - name: ""
         consumer_key: ""
         consumer_secret: ""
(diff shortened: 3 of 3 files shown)


lindhe:
Regarding the renaming of `extra_config` to `extraConfig`: I think this is a good change, and I wish we did the same for all the snake case values (but not in this PR).

Printing a deprecation warning like you do is very helpful, so even if we're still on major version zero maybe it's nice to keep there to give people some time to migrate away from it. I think we can keep that warning and remove the old value at the release of version `1.0.0`. Do you agree?

charts/invenio/templates/NOTES.txt
```diff
@@ -1 +1,11 @@
-Invenio is ready to rock 🚀
\ No newline at end of file
+Invenio is ready to rock 🚀
+
+Your release is named {{ .Release.Name }}.
+
+{{ if .Values.invenio.extra_config}}
```
lindhe, line 5:
```suggestion
{{- if .Values.invenio.extra_config }}
```

charts/invenio/templates/NOTES.txt
```diff
@@ -1 +1,11 @@
...
+
+DEPRECATION WARNING: 
+    `invenio.extra_config` has been renamed to `invenio.extraConfig` and will be
+    removed in a future release.
+
+{{ end }}
```
lindhe, line 11:
```suggestion
{{- end }}
```

charts/invenio/templates/invenio-configmap.yaml
```diff
@@ -20,5 +20,8 @@ data:
   INVENIO_DATACITE_ENABLED: "False"
   INVENIO_LOGGING_CONSOLE_LEVEL: "WARNING"
   {{- range $key, $value := .Values.invenio.extra_config }}
-  {{ $key }}: {{ $value | toYaml | indent 4 | trim }}
+  {{ $key }}: {{ tpl $value $ | toYaml | indent 5 | trim }}
```
lindhe, line 23: I'm leaning towards that it's OK to run the `extraConfig` values through `tpl`. I'm guessing there won't be any issues with that, especially since you have documented that the values are templated.

But if we are so careful with the old `extra_config` value as to print a deprecation warning, then I think we should not change its behavior at all.

So basically:
- I'm good with templating the strings from `extraConfig` (unless someone can show me an example of where it breaks).
- Let's keep the `extra_config` values non-templated.

Does this work for you?

egabancho: That makes sense.

egabancho:
My idea behind the renaming and the deprecation warnings is to do as many as we can with the changes/improvements we are making until we do the `1.0.0` release. For example, if we're working on a particular key that needs renaming, we will do it there and then.
Before the `1.0.0` release, I'd imagine we can make a bigger push and fix the rest, along with everything else that might need some TLC.
I don't have a strong opinion about when the deprecations should be removed. I guess we will cross that bridge when the time comes 😉

charts/invenio/templates/invenio-configmap.yaml
```diff
@@ -20,5 +20,8 @@ data:
   INVENIO_DATACITE_ENABLED: "False"
   INVENIO_LOGGING_CONSOLE_LEVEL: "WARNING"
   {{- range $key, $value := .Values.invenio.extra_config }}
-  {{ $key }}: {{ $value | toYaml | indent 4 | trim }}
+  {{ $key }}: {{ $value $ | toYaml | indent 5 | trim }}
```
lindhe, line 23: I think you missed removing the `$` here when you removed `tpl`

charts/invenio/templates/invenio-configmap.yaml
```diff
@@ -20,5 +20,8 @@ data:
...
+  {{- end }}
+  {{- range $key, $value := .Values.invenio.extraConfig }}
+  {{ $key }}: {{ tpl $value $ | toYaml | indent 5 | trim }}
```
lindhe, line 26: I have not tested this myself, so maybe it's obvious. But why `indent 5`? I think it's more common to keep YAML indentation at even multiples of 2 or 4. Am I missing something?

charts/invenio/templates/invenio-configmap.yaml
egabancho: Good catch!

charts/invenio/templates/invenio-configmap.yaml
egabancho: Honestly, I don't know. It was probably me testing the template part ☺️ 
It does render correctly though ... or at least it renders the same output with both values ...
```
$ helm template -f test.yaml  ./charts/invenio -s templates/invenio-configmap.yaml
---
# Source: invenio/templates/invenio-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: invenio-config
data:
  INVENIO_ACCOUNTS_SESSION_REDIS_URL: 'redis://....'
  INVENIO_APP_ALLOWED_HOSTS: '["test.com"]'
  ...
  INVENIO_CACHE_KEY_PREFIX: 'test::cache::'
  INVENIO_MAIL_DEFAULT_SENDER: no-reply@test.com
  ...
```

egabancho: I put it back to `4`, like it was originally.

lindhe (approved):

lindhe (changes requested):
Wait! I change my approval! We should bump Chart version also!

egabancho:
Do you think we can lump together a few of the changes and then do the version bump? I don't mind waiting for my changes to get released 😜

lindhe (approved):

PR #134, opened 2024-12-02
chore: add chart releaser action and increase version

### LINKED ISSUE #133: Publish the chart using github pages
It'd be nice to pull this chart directly without cloning the repository. It seems that using GitHub pages is a sensible solution, even document on helm's docs: https://helm.sh/docs/howto/chart_releaser_action/

I know we will probably make breaking changes, always for the better, but that's what version [`0.x.y`](https://semver.org/#spec-item-4) is for, although we will try to be nice.

cc-ing @lindhe

egabancho:
closes #133

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release Charts
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.6.0
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - master
```
egabancho, line 6: Should we do this on every commit or tags only?

Samk13: I would say Tags, as it keeps releases intentional. Triggering on every commit offers faster updates but risks noise and less meaningful versions.  I am open to other opinions as well!

lindhe: While I generally favor solutions that are fully automated, I think it might be a good idea to use tags for releases. Otherwise, we must somehow ensure that each commit also includes a version bump. And if we handle it well, I think it would be easier for people to find a specific version of the code matching the chart version they use.

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
```
lindhe, line 12: In my opinion, we should not use latest. It's so annoying when latest change and creates new behavior for seemingly no reason.

.github/workflows/release.yml
egabancho: I agree. I only put it to master because that's what they have in the  docs ☺️

.github/workflows/release.yml
egabancho: Totally. I'll pin it to the same version Invenio Github actions are using, `ubuntu-20.04`, https://github.com/inveniosoftware/.github/blob/master/workflow-templates/pypi-publish.yml#L10

Samk13 (approved):

PR #137, opened 2024-12-03
Enable custom vocabularies via ConfigMaps

### LINKED ISSUE #135: Allow passing vocabulary files to the pods
We have several instances that share the same docker image. It'd be nice to configure the vocabularies each one loads using the helm chart.

Maybe something like this:
```
invenio:
  ...
  vocabularies:
    resource_types.yaml: | 
        - id: publication
          icon: file alternate
          props:
            csl: report
            datacite_general: Text
            datacite_type: ""   
            openaire_resourceType: "0017"
            openaire_type: publication
            eurepo: info:eu-repo/semantics/other
            schema.org: https://schema.org/CreativeWork
            subtype: ""
            type: publication
            marc21_type: publication
            marc21_subtype: ""
          title:
            en: Publication
          tags:
            - depositable
            - linkable
          ...
```

egabancho:
* Adds the option to customize vocabularies (closes #135)
* Adds proper labels everywhere

### DIFF
--- a/charts/invenio/templates/install-init-job.yaml
+++ b/charts/invenio/templates/install-init-job.yaml
@@ -57,10 +57,22 @@ spec:
             secretKeyRef:
               name: invenio-secrets
               key: INVENIO_CSRF_SECRET_SALT
+        volumeMounts:
+          {{- range $key, $value := .Values.invenio.vocabularies }}
+          - name: vocabularies
+            mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+            subPath: "{{ $key }}"
+          {{- end }}
         {{- if .Values.web.resources }}
         resources: {{- toYaml .Values.web.resources | nindent 10 }}
         {{- end }}
       restartPolicy: OnFailure
+      {{- if .Values.invenio.vocabularies }}
+      volumes:
+        - name: vocabularies
+          configMap:
+            name: "invenio-vocabularies"
+      {{- end }}
       {{- if .Values.web.imagePullSecret }}
       imagePullSecrets:
       - name: {{ .Values.web.imagePullSecret }}
--- a/charts/invenio/templates/vocabularies-configmap.yaml
+++ b/charts/invenio/templates/vocabularies-configmap.yaml
@@ -1,0 +1,10 @@
+{{- with .Values.invenio.vocabularies }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: invenio-vocabularies
+  labels:
+    {{- include "invenio.labels" $ | nindent 4 }}
+data:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
--- a/charts/invenio/templates/web-deployment.yaml
+++ b/charts/invenio/templates/web-deployment.yaml
@@ -131,6 +131,11 @@ spec:
           - name: kerberos-credentials-cache
             mountPath: /tmp
           {{- end }}
+          {{- range $key, $value := .Values.invenio.vocabularies }}
+          - name: vocabularies
+            mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+            subPath: "{{ $key }}"
+          {{- end }}
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
@@ -278,6 +283,11 @@ spec:
       {{- if .Values.kerberos.enabled }}
       - name: kerberos-credentials-cache
         emptyDir: {}
+      {{- end }}
+      {{- if .Values.invenio.vocabularies }}
+      - name: vocabularies
+        configMap:
+          name: "invenio-vocabularies"
       {{- end }}
       {{- if .Values.web.imagePullSecret }}
       imagePullSecrets:
--- a/charts/invenio/templates/worker-beat-deployment.yaml
+++ b/charts/invenio/templates/worker-beat-deployment.yaml
@@ -35,6 +35,11 @@ spec:
         - mountPath: /opt/invenio/var/instance/data
           name: shared-volume
         {{- end }}
+        {{- range $key, $value := .Values.invenio.vocabularies }}
+        - name: vocabularies
+          mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+          subPath: "{{ $key }}"
+        {{- end }}
         envFrom:
         - configMapRef:
             name: invenio-config
@@ -126,6 +131,11 @@ spec:
       - name: shared-volume
         persistentVolumeClaim:
           claimName: {{ .Values.persistence.name }}
+      {{- end }}
+      {{- if .Values.invenio.vocabularies }}
+      - name: vocabularies
+        configMap:
+          name: "invenio-vocabularies"
       {{- end }}
       {{- if .Values.worker.imagePullSecret }}
       imagePullSecrets:
--- a/charts/invenio/templates/worker-deployment.yaml
+++ b/charts/invenio/templates/worker-deployment.yaml
@@ -114,6 +114,11 @@ spec:
             - name: kerberos-credentials-cache
               mountPath: /tmp
             {{- end }}
+            {{- range $key, $value := .Values.invenio.vocabularies }}
+            - name: vocabularies
+              mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+              subPath: "{{ $key }}"
+            {{- end }}
         {{- if .Values.kerberos.enabled }}
         - name: kerberos-credentials
           image: {{ required "Missing .Values.kerberos.image" .Values.kerberos.image }}
@@ -165,7 +170,11 @@ spec:
         - name: kerberos-credentials-cache
           emptyDir: {}
         {{- end }}
-
+        {{- if .Values.invenio.vocabularies }}
+        - name: vocabularies
+          configMap:
+            name: "invenio-vocabularies"
+        {{- end }}
       {{- if .Values.worker.imagePullSecret }}
       imagePullSecrets:
       - name: {{ .Values.worker.imagePullSecret }}
--- a/charts/invenio/values.yaml
+++ b/charts/invenio/values.yaml
@@ -49,6 +49,31 @@ invenio:
     #       name: name-of-my-secret
     #       key: KEY_IN_MY_SECRET
 
+  ## @param invenio.vocabularies Vocabularies to be loaded as files under /app_data/vocabularies
+  ## Example
+  ## vocabularies:
+  ##   resource_types.yaml: |
+  ##    - id: publication
+  ##      icon: file alternate
+  ##      props:
+  ##        csl: report
+  ##        datacite_general: Text
+  ##        datacite_type: ""
+  ##        openaire_resourceType: "0017"
+  ##        openaire_type: publication
+  ##        eurepo: info:eu-repo/semantics/other
+  ##        schema.org: https://schema.org/CreativeWork
+  ##        subtype: ""
+  ##        type: publication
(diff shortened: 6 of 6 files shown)


lindhe:
Can vocabularies really fit in a ConfigMap? I was under the impression that they are often pretty big.

charts/invenio/templates/logstash/filebeat-configmap.yaml
```diff
@@ -4,6 +4,8 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: filebeat-config
+  labels:
```
lindhe, line 7: I like that you try to fix the label situation, but I think this belongs in another PR. It's hard to review multiple things in one PR on GitHub.

lindhe: But again: I _really_ appreciate how good you are at creating separate commits for each change. It's so nice to see someone that creates atomic commits for a change! ❤️

egabancho: I know; it's just faster for me to work like that sometimes. I'm making some of the changes because I feel they need to be done, not because we actually need them.

On the `Files changed` tab, you can select the commit you want to check. That may help.

egabancho: Also, I just realized that I based this one on top of the other PR ... that's probably also confusing 😂

lindhe: Yeah, the changes that are from other PRs must for sure be removed here, very hard to review otherwise. 🙂

egabancho: Done, I moved the label stuff [here](https://github.com/inveniosoftware/helm-invenio/pull/138) and rebase.

charts/invenio/templates/install-init-job.yaml
```diff
@@ -56,10 +56,24 @@ spec:
             secretKeyRef:
               name: invenio-secrets
               key: INVENIO_CSRF_SECRET_SALT
+        volumeMounts:
+          {{- if .Values.invenio.vocabularies }}
```
lindhe, line 60: I think this if is redundant or misplaced. I have not run it myself to verify, but I think `range` will run 0 times (not error) if `.Values.invenio.vocabularies` is null or has no elements.

On the other hand, if `.Values.invenio.vocabularies` is null then you will still have the `volumeMounts` field printed with empty contents (since the `volumeMounts` is defined outside of the `if`).

charts/invenio/templates/install-init-job.yaml
```diff
@@ -56,10 +56,24 @@ spec:
...
+            mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+            subPath: "{{ $key }}"
+            {{- end }}
+          {{- end }}
         {{- if .Values.web.resources }}
         resources: {{- toYaml .Values.web.resources | nindent 10 }}
         {{- end }}
       restartPolicy: OnFailure
+      volumes:
+        {{- if .Values.invenio.vocabularies}}
```
lindhe, line 72: If `.Values.invenio.vocabularies` is null, `volumes` will be empty. Better to not include `volumes` if not used?

charts/invenio/templates/vocabularies-configmap.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: ConfigMap
+metadata:
+  name: invenio-vocabularies
+  labels:
+    {{- include "invenio.labels" . | nindent 4 }}
+data:
+{{- range $key, $value := .Values.invenio.vocabularies }}
+  {{ $key }}: |
+  {{- $value | nindent 4 }}
+{{- end }}
```
lindhe, line 12: Not a bug, I just want to highlight that this logic can probably be replaced by `{{- .Values.invenio.vocabularies | toYaml | nindent 2 }}` or something like that. No need to iterate over the objects.

charts/invenio/templates/web-deployment.yaml
```diff
@@ -126,6 +126,13 @@ spec:
           - name: kerberos-credentials-cache
             mountPath: /tmp
           {{- end }}
+          {{- if .Values.invenio.vocabularies }}
```
lindhe, line 129: I think this `if` is redundant.

charts/invenio/templates/worker-beat-deployment.yaml
```diff
@@ -33,6 +33,13 @@ spec:
         - mountPath: /opt/invenio/var/instance/data
           name: shared-volume
         {{- end }}
+        {{- if .Values.invenio.vocabularies }}
```
lindhe, line 36: I think this `if` is redundant.

charts/invenio/templates/worker-deployment.yaml
```diff
@@ -109,6 +109,13 @@ spec:
             - name: kerberos-credentials-cache
               mountPath: /tmp
             {{- end }}
+            {{- if .Values.invenio.vocabularies }}
```
lindhe, line 112: I think this `if` is redundant.

charts/invenio/templates/vocabularies-configmap.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.invenio.vocabularies}}
```
lindhe, line 1:
```suggestion
{{- if .Values.invenio.vocabularies }}
```

charts/invenio/templates/vocabularies-configmap.yaml
```diff
@@ -0,0 +1,8 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: invenio-vocabularies
+data:
+{{- .Values.invenio.vocabularies | toYaml | nindent 2 }}
+{{- end }}
```
lindhe, line 8:
```suggestion
{{- with .Values.invenio.vocabularies }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: invenio-vocabularies
data:
  {{- toYaml . | nindent 2 }}
{{- end }}
```

lindhe: Sorry for being nit-picky. But I like reducing code repetition. It's so easy to make typos in the future if multiple places needs to be changed.

charts/invenio/templates/web-deployment.yaml
```diff
@@ -274,6 +279,11 @@ spec:
       - name: kerberos-credentials-cache
         emptyDir: {}
       {{- end }}
+      {{- if .Values.invenio.vocabularies}}
```
lindhe, line 282:
```suggestion
      {{- if .Values.invenio.vocabularies }}
```

charts/invenio/templates/worker-beat-deployment.yaml
```diff
@@ -122,6 +127,11 @@ spec:
         persistentVolumeClaim:
           claimName: {{ .Values.persistence.name }}
       {{- end }}
+      {{- if .Values.invenio.vocabularies}}
```
lindhe, line 130:
```suggestion
      {{- if .Values.invenio.vocabularies }}
```

charts/invenio/templates/worker-deployment.yaml
```diff
@@ -160,7 +165,11 @@ spec:
         - name: kerberos-credentials-cache
           emptyDir: {}
         {{- end }}
-
+        {{- if .Values.invenio.vocabularies}}
```
lindhe, line 168:
```suggestion
        {{- if .Values.invenio.vocabularies }}
```

charts/invenio/templates/install-init-job.yaml
```diff
@@ -56,10 +56,22 @@ spec:
...
+          {{- range $key, $value := .Values.invenio.vocabularies }}
+          - name: vocabularies
+            mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+            subPath: "{{ $key }}"
+          {{- end }}
         {{- if .Values.web.resources }}
         resources: {{- toYaml .Values.web.resources | nindent 10 }}
         {{- end }}
       restartPolicy: OnFailure
+      {{- if .Values.invenio.vocabularies}}
```
lindhe, line 69:
```suggestion
      {{- if .Values.invenio.vocabularies }}
```

charts/invenio/templates/install-init-job.yaml
```diff
@@ -56,10 +56,22 @@ spec:
             secretKeyRef:
               name: invenio-secrets
               key: INVENIO_CSRF_SECRET_SALT
+        volumeMounts:
+          {{- range $key, $value := .Values.invenio.vocabularies }}
+          - name: vocabularies
+            mountPath: "/opt/invenio/var/instance/app_data/vocabularies/{{ $key }}"
+            subPath: "{{ $key }}"
```
lindhe, line 63: Is `/opt/invenio/var/instance/app_data/vocabularies/` generally prepopulated by the image, so other files exists there without mount?

If so, we do need to use `subPath` like you do here, so the other files are not overwritten by the volume mount.

On the other hand, if that path is not prepopulated by the image, then I it would be advantageous to drop `subPath`. Mostly because `subPath` mounts are not updated without a pod restart, but also because that would let us drop the entire `range` since each field in a `ConfigMap` gets mounted then.

(same comment for the similar places in this PR).

lindhe (changes requested):
Sorry for being so slow to review. I hate being a blocker.

Just have some minor comments here, mostly style changes. As soon as they are fixed (or answered) I approve this PR. 👍

charts/invenio/templates/vocabularies-configmap.yaml
egabancho: You **have to be** nit-picky, that's what we pay you for 😜

charts/invenio/templates/install-init-job.yaml
egabancho: In general, the folder already existing inside the default docker image, https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/Dockerfile#L20 and https://github.com/inveniosoftware/cookiecutter-invenio-rdm/tree/master/%7B%7Bcookiecutter.project_shortname%7D%7D/app_data/vocabularies.
I say _in general_ because one can always change the docker image and not add it ...

egabancho: It is more or less hardcoded here https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/__init__.py#L23

charts/invenio/templates/vocabularies-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+{{- with .Values.invenio.vocabularies }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: invenio-vocabularies
+  labels:
+    {{- include "invenio.labels" . | nindent 4 }}
```
lindhe, line 7: Please verify by running `helm template` if `include "invenio.labels" .` works or if we need to use `include "invenio.labels" $` instead (since `with` changes the scope). I'm not sure, we need to check.

charts/invenio/templates/install-init-job.yaml
lindhe: Thanks for checking. Then I think we must use `subPath`. Thus, this part looks good in its current form.

charts/invenio/templates/vocabularies-configmap.yaml
egabancho: Good catch!

lindhe (approved):

PR #138, opened 2024-12-05
chore: add proper labels

### LINKED ISSUE #114: Default and unique label sets that comply with the k8s specifications
All invenio k8s resources (deployment, service, etc) should have a unique label set for common and selector labels which can be assigned to a helm release and matches the k8s conventions.

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Possible solution:

These label sets are also generated by `helm create...` by default:

```yaml
{{/*
Common labels
*/}}
{{- define "invenio.labels" -}}
helm.sh/chart: {{ include "invenio.chart" . }}
{{ include "invenio.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{/*
Selector labels
*/}}
{{- define "invenio.selectorLabels" -}}
app.kubernetes.io/name: {{ include "invenio.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}
```

egabancho:
* closes #114

### DIFF
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -1,3 +1,37 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "invenio.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "invenio.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "invenio.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "invenio.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "invenio.labels" -}}
+helm.sh/chart: {{ include "invenio.chart" . }}
+{{ include "invenio.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
 ###########################     Invenio hostname     ###########################
 {{/*
   This template renders the hostname for Invenio.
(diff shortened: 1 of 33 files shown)


charts/invenio/templates/_helpers.tpl
```diff
@@ -1,3 +1,53 @@
...
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+Selector labels
+*/}}
+{{- define "invenio.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "invenio.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
lindhe, line 49: I think it's valid to define a template after it's used (it's used on line 38 and defined here at line 46). But I think it reads better for people if we define it first. Also it's missing a blank newline.

charts/invenio/templates/_helpers.tpl
```diff
@@ -1,3 +1,53 @@
...
+{{- define "invenio.labels" -}}
+helm.sh/chart: {{ include "invenio.chart" . }}
+{{ include "invenio.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+Selector labels
+*/}}
```
lindhe, line 45: Missing start of block comment: `{{*/`

charts/invenio/templates/_helpers.tpl
```diff
@@ -1,3 +1,53 @@
...
+{{- define "invenio.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "invenio.fullname" -}}
```
lindhe, line 13: Seems like you define `invenio.fullname` but never use it.

charts/invenio/templates/_helpers.tpl
egabancho: I just ran `helm create invenio` on a different folder and copy over the content from `_helpers.tpl` ☺️

charts/invenio/templates/_helpers.tpl
egabancho: Removed.

lindhe (approved):

PR #144, opened 2024-12-12
Nginx updates

egabancho:
* nginx: add "application/javascript" MIME type for mjs files
* nginx: disable chunked transfer encoding for proxied apps

Kudos to @max-moser for doing all the work. I'm just porting it here 😉


lindhe (commented):
I have no way to testing this, but I assume it's good. 🙂

lindhe (approved):
I have no way to testing this, but I assume it's good. 🙂

PR #146, opened 2024-12-13
pods: add extraEnvVars to all invenio pods

### LINKED ISSUE #136: Allow custom configuration variables for web and worker pods
It'll be interesting to allow to have different configurations depending on the kind of deployment. 
I'm thinking, in particular, setting SENTRY_TRANSPORT = 'raven.transport.http.HTTPTransport' on the celery workers https://invenio-logging.readthedocs.io/en/latest/configuration.html.

I'm thinking something as simple as:
```
worker:
  extraEnvVars:
    - name: INVENIO_SENTRY_TRANSPORT
      value: 'raven.transport.http.HTTPTransport' 
```

And then render them as environmental variables directly on the container.

egabancho:
* Closes #136

### DIFF
--- a/charts/invenio/templates/web-deployment.yaml
+++ b/charts/invenio/templates/web-deployment.yaml
@@ -39,6 +39,9 @@ spec:
           value: {{ include "invenio.rabbitmq.uri" . }}
         - name: INVENIO_SQLALCHEMY_DATABASE_URI
           value: {{ include "invenio.sqlAlchemyDbUri" . }}
+        {{- with .Values.web.extraEnvVars }}
+        {{-  toYaml . | nindent 8 }}
+        {{- end }}
         - name: INVENIO_SECRET_KEY
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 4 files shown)


charts/invenio/templates/web-deployment.yaml
```diff
@@ -37,6 +37,9 @@ spec:
           value: {{ include "invenio.rabbitmq.uri" . }}
         - name: INVENIO_SQLALCHEMY_DATABASE_URI
           value: {{ include "invenio.sqlAlchemyDbUri" . }}
+        {{- if .Values.web.extraEnvVars }}
+        {{- .Values.web.extraEnvVars | toYaml | nindent 8 }}
+        {{- end }}
```
lindhe, line 42: To reduce code duplication, I prefer using `with` statement:

```suggestion
        {{- with .Values.web.extraEnvVars }}
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

But if-statements like this are clearly valid too, so I think we can go ahead with this. Just wanted to mention it. 🙂

lindhe (approved):
LGTM. Good work! :)

My comment does not need to be fixed unless you deem it worthwhile.

lindhe (approved):

PR #151, opened 2025-01-13
Give resources unique names and improve labels

egabancho:


egabancho:
@lindhe, for the non-configurable names I just prepended `invenio-fullname` every place they occur. Do you think we should have a tpl function for that?

charts/invenio/templates/_helpers.tpl
```diff
@@ -5,6 +5,24 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
```
lindhe, line 11:
```suggestion
If release name contains chart name, the release name will be used as a full name.
```

charts/invenio/templates/_helpers.tpl
```diff
@@ -283,7 +301,7 @@ Get the sentry secret name
 {{- else if  .Values.invenio.sentry.secret_name -}}
   {{- print .Values.invenio.sentry.secret_name -}}  
 {{- else -}}
-  {{- "sentry-secrets" -}}
+  {{- printf "%s-%s" (include "invenio.fullname" .) "sentry-secrets" -}}
```
lindhe, line 304: A comment I think holds true for multiple places:

Your change looks good. But I dislike the old name `sentry-secrets` that we keep. I think having the resource kind in the resource name is redundant; in my experience it's always clear from context or metadata if it's a secret or some other object. I think it looks nicer to have a Secret called "invenio-sentry" than a secret called "invenio-sentry-secrets".

What do you think?

lindhe:
What's with 8fb7803193660e75456fe1e04083de0121744ed4 ?

Does that belong in this PR? If so: I don't understand the change.

lindhe:
> @lindhe, for the non-configurable names I just prepended `invenio.fullname` every place they occur. Do you think we should have a tpl function for that? Similar to what `invenio.secretName` does, but simpler ...

I think it's better to add that later, if/when there's a need for it, instead of increasing the complexity already. But I'm open for it if you think it would be a good idea.

charts/invenio/templates/_helpers.tpl
egabancho: I just copy paste the output of `helm create`, didn't even read it 😂

egabancho:
> What's with [8fb7803](https://github.com/inveniosoftware/helm-invenio/commit/8fb7803193660e75456fe1e04083de0121744ed4) ?
> 
> Does that belong in this PR? If so: I don't understand the change.

I does to some degree. I forgot to remove that bit when I refactor the secrets. The are all loaded [here](https://github.com/inveniosoftware/helm-invenio/blob/master/charts/invenio/templates/web-deployment.yaml#L33) rather then individually like before.
If I don't these lines, I'll have to update the name of the secret accordingly to later remove them becuase I forgot to do it elarlier ... sounds like a bit of waste of time to me.

Does it make sense now? I can change the commit message to something more meaningful if you want.

charts/invenio/templates/_helpers.tpl
egabancho: I think the sentry secret and the ingress are the only resources that include the resource type in the name actually.
The secret that contains the secret keys and salts is called `invenio.fullname-secrets`, but I am inclined to think this is correct, I don't know what else we could call it.

lindhe: Yeah, it's very hard to know whether to make yet another improvement or keep the change simpler. I'm not sure either, but if this occurred to both of us then that's probably a sign that it might be worthwhile.

lindhe:
> Does it make sense now? I can change the commit message to something more meaningful if you want.

Yeah, we can keep it as long as the commit messages describe what's happening. 👍

charts/invenio/templates/_helpers.tpl
lindhe: > The secret that contains the secret keys and salts is called `invenio.fullname-secrets`, but I am inclined to think this is correct, I don't know what else we could call it.

I'm inclined to say that it should simply be called `invenio.fullname`. Same logic as before; I already know it's a Secret. But it's fine by me if you want to keep the `-secrets` suffix, there's clearly no harm to it.

lindhe: I think everything else here looks great, so just tell me if you want to keep the suffix or not and then I'll approve.

egabancho: I changed the secret name to just `invenio.fullname`, you convince me 😅

lindhe (approved):
Great work, as always! 👍

Issue #112, opened 2024-03-24
Granular env-based solution for "connection string"-like config


slint:
It's common practice to be able to configure services that rely on connection strings/URIs (e.g. DB, OpenSearch, RabbitMQ, Redis) by individually setting parts of the string via env variables. This allows:

- reusing config in sub-charts (e.g. passing `RABBITMQ_USER` to both the `web`/`worker` config, but also the RabbitMQ sub-chart)
- placing only sensitive information of the config in a secret and exposing as an env var

On the application-side building the `SQLALCHEMY_DATABASE_URI` config would look something like:

```python
import os

# First check if the full value is set
if os.environ.get("SQLALCHEMY_DATABASE_URI"):
    SQLALCHEMY_DATABASE_URI = os.environ.get("SQLALCHEMY_DATABASE_URI")
else:
    # Check parts of the config
    db_host = os.environ.get("DB_HOST")
    db_password = os.environ.get("DB_PASSWORD")
    db_user = os.environ.get("DB_USER")
    db_name = os.environ.get("DB_NAME")
    db_port = os.environ.get("DB_PORT")

    if all([db_host, db_password, db_user, db_name, db_port]):
        SQLALCHEMY_DATABASE_URI = f"postgresql://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}"
    else:
        SQLALCHEMY_DATABASE_URI = f"postgresql://invenio:password@localhost:5432/invenio"
```

> [!IMPORTANT]
> This requires development in the application modules that configure services like the DB, OpenSearch, RabbitMQ, etc. so that they support loading this type of config.

lindhe:
Ah, yes! This is an important improvement, in my opinion!

An adjacent topic to this is if we can provide credentials using secrets! It would be a great improvement on security if we could set all credentials using secretes (which is not possible today, if I recall correctly). We may also want to consider using credentials from the secrets generated by our chart dependencies.

I think these topics should be part of this issue, but if you think otherwise I can create a separate issue for them.

slint:
@lindhe exactly, I tried to (badly 😅) summarize this in:

> placing only sensitive information of the config in a secret and exposing as an env var

We touched a bit on this [on Discord](https://discord.com/channels/692989811736182844/1179090710402715770/1205086647864135770), and I brought it up at the InvenioRDM workshop last week, where people agreed based on their experience with other Helm Charts approach to secrets.

PR #152, opened 2025-01-14
Better connection strings

egabancho:


charts/invenio/templates/_helpers.tpl
```diff
@@ -199,79 +199,110 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
+{{- if or (and .Values.postgresql.enabled .Values.postgresql.auth.password) .Values.postgresqlExternal.password }}
+  value: {{ include "invenio.postgresql.password" .  | quote }}
+{{- else }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ include "invenio.postgresql.secretName" .}}
+      key: {{ include "invenio.postgresql.secretKey" .}}
+{{- end }}
+- name: INVENIO_SQLALCHEMY_DATABASE_URI
+  value: "$(INVENIO_DB_PROTOCOL)://$(INVENIO_DB_USER):$(INVENIO_DB_PASSWORD)@$(INVENIO_DB_HOST):$(INVENIO_DB_PORT)/$(INVENIO_DB_NAME)"
```
egabancho, line 305: @lindhe What do you think about something like this? It should somehow address https://github.com/inveniosoftware/helm-invenio/issues/112

lindhe: Wow. Maybe. 👀 The logic with defining each part as a separate env is sound, no issues there. But composing a new env value based on other envs like this, I'm unfamiliar with. If it works as your code indicates, then yeah I think this looks very promising.

How does this work? Does Kubernetes support the `$(ENV)` syntax in string values, or does that happen inside the container? I feel lost as to why this works. 😅 I assumed you tried it and it seems to work?

Assuming it works, I really like this because I think it should be forwards-compatible with https://github.com/inveniosoftware/invenio-config/pull/57 and/or https://github.com/inveniosoftware/invenio-app-rdm/pull/2918 which is nice.

egabancho: I am not sure when it happens, in particular after reading this in the docs `Variables making use of others defined in the same context must come later in the list.`.
I would think the interpolation happens at run time inside the container. This `Environment` section of the output of `kubectl describe pod` 
```
     Environment:
      SECRET:   <set to the key 'secret' in secret 'compose-env'>  Optional: false
      PLAIN:    plain
      MESSAGE:  $(PLAIN) - $(SECRET)
```

lindhe: Sounds convincing! Wow, this is so powerful (and a bit confusing)! 😀 

I think this would be a great first step towards handling connection credentials better!

plesoun-stein:
we have cloudnative postgres operator and all params are in secret. 
It should be nice to specify `password` from secret

helper should look  like
```
{{- define "invenio.postgresql.password" -}}
  {{- if .Values.postgresql.enabled -}}
    {{- required "Missing .Values.postgresql.auth.password" .Values.postgresql.auth.password -}}
  {{- else if and ( not .Values.postgresql.enabled ) ( not ( empty .Values.postgresqlExternal.password )) -}}
    {{- .Values.postgresqlExternal.password -}}
  {{- else if and ( not .Values.postgresql.enabled ) ( empty .Values.postgresqlExternal.password ) ( not ( empty  .Values.postgresqlExternal.existingSecret )) }}
    {{- $obj := (lookup "v1" "Secret" $root.Release.Namespace .Values.postgresqlExternal.existingSecret ).data }}
    {{- if or ( empty $obj ) ( empty ( get $obj .Values.postgresqlExternal.existingSecretPasswordKey )) -}}
      {{- fail ( printf "\n\n\tthere is missing secret \`%s\` or password key \`%s\`" .Values.postgresqlExternal.existingSecret .Values.postgresqlExternal.existingSecretPasswordKey ) }}
    {{- end -}}
    {{- printf "%s" ( get $obj ( tpl .Values.postgresqlExternal.existingSecretPasswordKey . ) | b64dec )  }}
  {{- else -}}
    {{- fail ( printf "\n\n\tInternal postgresql chart is disabled and there is missing .Values.postgresqlExternal.existingSecret declaration or another mistake or typo\n\n%s" ( toYaml .Values.postgresqlExternal | indent 6 )) }}
  {{- end -}}
{{- end -}}
```
(it's not tested, only example as POC) 
simmilar work may be done for other params. hostname, username etc. 
and I cat write it  reusable for rabbit and redis

I have a time and I can do it.

charts/invenio/templates/_helpers.tpl
```diff
@@ -217,79 +235,110 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
 {{/*
   This template renders the port number used for the PostgreSQL instance.
 */}}
 {{- define "invenio.postgresql.port" -}}
   {{- if .Values.postgresql.enabled -}}
-    {{- required "Missing .Values.postgresql.primary.service.ports.postgresql" .Values.postgresql.primary.service.ports.postgresql -}}
+    {{- required "Missing .Values.postgresql.primary.service.ports.postgresql" (tpl .Values.postgresql.primary.service.ports.postgresql .) -}}
   {{- else -}}
-    {{- required "Missing .Values.postgresqlExternal.port" .Values.postgresqlExternal.port -}}
+    {{- required "Missing .Values.postgresqlExternal.port" (tpl .Values.postgresqlExternal.port .) -}}
```
lindhe, line 302: I tried installing this and got the following error:

```text
Error: INSTALLATION FAILED: template: invenio/templates/worker-deployment.yaml:43:14: executing "invenio/templates/worker-deployment.yaml" at <include "invenio.config.database" .>: error calling include: template: invenio/templates/_helpers.tpl:326:12: executing "invenio.config.database" at <include "invenio.postgresql.port" .>: error calling include: template: invenio/templates/_helpers.tpl:300:91: executing "invenio.postgresql.port" at <.Values.postgresql.primary.service.ports.postgresql>: wrong type for value; expected string; got float64
```

Guessing this is where it fails?

plesoun-stein: You can try quote port number, 
row 326
`value: {{ include "invenio.postgresql.port" . }}`   
expects string as it renders template.

row 302
convert to string is better.  
`    {{- required "Missing .Values.postgresqlExternal.port" ((tpl .Values.postgresqlExternal.port .) | toString) -}}

lindhe: The issue was that the `tpl` function takes a string parameter, but the port was a `float64`. Have fixed it now.

charts/invenio/templates/_helpers.tpl
lindhe: As far as I can tell, this works now! 🚀

egabancho:
@lindhe, Take another look 😉

charts/invenio/templates/_helpers.tpl
```diff
@@ -119,43 +117,61 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
 {{/*
   This template renders the AMQP port number for RabbitMQ.
 */}}
 {{- define "invenio.rabbitmq.amqpPort" -}}
   {{- if .Values.rabbitmq.enabled }}
-    {{- required "Missing .Values.rabbitmq.service.ports.amqp" .Values.rabbitmq.service.ports.amqp -}}
+    {{- required "Missing .Values.rabbitmq.service.ports.amqp" .Values.rabbitmq.service.ports.amqp | quote -}}
   {{- else }}
-    {{- required "Missing .Values.rabbitmqExternal.amqpPort" .Values.rabbitmqExternal.amqpPort -}}
+    {{- required "Missing .Values.rabbitmqExternal.amqpPort" (tpl .Values.rabbitmqExternal.amqpPort .) | quote -}}
```
lindhe, line 149:
```suggestion
    {{- required "Missing .Values.rabbitmqExternal.amqpPort" (tpl (toString .Values.rabbitmqExternal.amqpPort) .) | quote -}}
```

Not sure which type `amqpPort` is, but if it can be a number we need to cast it to string for `tpl` to work.

charts/invenio/templates/_helpers.tpl
```diff
@@ -119,43 +117,61 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
 {{/*
   This template renders the management port number for RabbitMQ.
 */}}
 {{- define "invenio.rabbitmq.managementPort" -}}
   {{- if .Values.rabbitmq.enabled }}
-    {{- required "Missing .Values.rabbitmq.service.ports.manager" .Values.rabbitmq.service.ports.manager -}}
+    {{- required "Missing .Values.rabbitmq.service.ports.manager" .Values.rabbitmq.service.ports.manager | quote -}}
   {{- else }}
-    {{- required "Missing .Values.rabbitmqExternal.managementPort" .Values.rabbitmqExternal.managementPort -}}
+    {{- required "Missing .Values.rabbitmqExternal.managementPort" (tpl .Values.rabbitmqExternal.managementPort .) | quote -}}
```
lindhe, line 160:
```suggestion
    {{- required "Missing .Values.rabbitmqExternal.managementPort" (tpl (toString .Values.rabbitmqExternal.managementPort .) | quote -}}
```

charts/invenio/templates/_helpers.tpl
```diff
@@ -167,42 +183,47 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
+{{- else }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ include "invenio.rabbitmq.secretName" .}}
+      key: {{ include "invenio.rabbitmq.secretKey" .}}
+{{- end }}
+- name: INVENIO_BROKER_URL
+  value: {{ $uri}}
+- name: INVENIO_CELERY_BROKER_URL
+  value: {{ $uri}}
```
lindhe, line 224:
```suggestion
- name: INVENIO_BROKER_URL
  value: {{ $uri }}
- name: INVENIO_CELERY_BROKER_URL
  value: {{ $uri }}
```

lindhe:
I think maybe my merge via GitHub went wrong, because something is badly broken right now. I've reset to your last commit. Checking things again now…

lindhe (changes requested):
There's a bunch of relatively minor comments here. But in my last test, I think the values looked pretty broken when I ran `env` in the container. I'll see if I can narrow down that issue while we're addressing my other comments here.

charts/invenio/values.yaml
```diff
@@ -373,7 +375,28 @@ rabbitmq:
...
+rabbitmqExternal:
+  username: invenio
+  password: ""
+  amqpPort: 5671
+  managementPort: 15672
+  hostname: ""
+  protocol: ""
+  vhost: ""
+  existingSecret: ""
+  existingSecretPasswordKey: "amqp-password"
```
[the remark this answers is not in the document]
lindhe: Also: I believe that `5672` is the normal AMQP port. I think `5671` is AMQPS.

charts/invenio/templates/_helpers.tpl
```diff
@@ -167,42 +183,47 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
+{{- else }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ include "invenio.rabbitmq.secretName" .}}
+      key: {{ include "invenio.rabbitmq.secretKey" .}}
+{{- end }}
+- name: INVENIO_BROKER_URL
+  value: {{ $uri }}
+- name: INVENIO_CELERY_BROKER_URL
+  value: {{ $uri }}
```
lindhe, line 224: Since the rendered chart contains the full `$uri` string, this looks a bit messy in my opinion:

image

WDYT about using `$(INVENIO_BROKER_URL)` instead, to make it explicit that they are the same?

```suggestion
- name: INVENIO_BROKER_URL
  value: {{ $uri }}
- name: INVENIO_CELERY_BROKER_URL
  value: $(INVENIO_BROKER_URL)
```

charts/invenio/templates/_helpers.tpl
lindhe: There's a typo here! I'll fix a proper variant instead.

Samk13 (dismissed):

lindhe:
It seems to work well, but I'm a bit confused with these variables:

```console
$ kubectl exec deploy/invenio-web -c web -- env |  grep RABBITMQ_PORT
INVENIO_RABBITMQ_PORT_25672_TCP=tcp://10.43.72.254:25672
INVENIO_RABBITMQ_PORT_4369_TCP_ADDR=10.43.72.254
INVENIO_RABBITMQ_PORT_25672_TCP_ADDR=10.43.72.254
INVENIO_RABBITMQ_PORT_15672_TCP_ADDR=10.43.72.254
INVENIO_RABBITMQ_PORT_15672_TCP_PROTO=tcp
INVENIO_RABBITMQ_PORT=tcp://10.43.72.254:5672
INVENIO_RABBITMQ_PORT_25672_TCP_PORT=25672
INVENIO_RABBITMQ_PORT_5672_TCP_ADDR=10.43.72.254
INVENIO_RABBITMQ_PORT_25672_TCP_PROTO=tcp
INVENIO_RABBITMQ_PORT_5672_TCP_PROTO=tcp
INVENIO_RABBITMQ_PORT_4369_TCP_PORT=4369
INVENIO_RABBITMQ_PORT_5672_TCP_PORT=5672
INVENIO_RABBITMQ_PORT_4369_TCP_PROTO=tcp
INVENIO_RABBITMQ_PORT_4369_TCP=tcp://10.43.72.254:4369
INVENIO_RABBITMQ_PORT_15672_TCP=tcp://10.43.72.254:15672
INVENIO_RABBITMQ_PORT_15672_TCP_PORT=15672
INVENIO_RABBITMQ_PORT_5672_TCP=tcp://10.43.72.254:5672
```

I guess they are just strangely named variables in the container, not something we've set here that gets messed up, right?

charts/invenio/templates/_helpers.tpl
```diff
@@ -217,79 +238,110 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
-  {{- $port := include "invenio.postgresql.port" . -}}
-  {{- $databaseName := include "invenio.postgresql.databaseName" . -}}
-  {{- printf "postgresql+psycopg2://%s:%s@%s:%v/%s" $username $password $hostname $port $databaseName -}}
+{{- define "invenio.config.database" -}}
+- name: INVENIO_DB_USER
+  value: {{ include "invenio.postgresql.username" . }}
+- name: INVENIO_DB_HOST
+  value: {{ include "invenio.postgresql.hostname" . }}
+- name: INVENIO_DB_PORT
+  value: {{ include "invenio.postgresql.port" . | quote }}
```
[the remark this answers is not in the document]
lindhe: I've done this. Please mark as resolved if you disagree, or else we'll revert commit 0604805

charts/invenio/templates/_helpers.tpl
```diff
@@ -217,79 +238,110 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
+  Define database connection env section. 
 */}}
-{{- define "invenio.sqlAlchemyDbUri" -}}
-  {{- $username := include "invenio.postgresql.username" . -}}
-  {{- $password := include "invenio.postgresql.password" . -}}
-  {{- $hostname := include "invenio.postgresql.hostname" . -}}
-  {{- $port := include "invenio.postgresql.port" . -}}
-  {{- $databaseName := include "invenio.postgresql.databaseName" . -}}
-  {{- printf "postgresql+psycopg2://%s:%s@%s:%v/%s" $username $password $hostname $port $databaseName -}}
+{{- define "invenio.config.database" -}}
```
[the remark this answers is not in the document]
lindhe: I see that `invenio.config.sentry` already exists and is used similarly to this template. While I'd prefer change the name, this PR is probably the wrong place to do that.

charts/invenio/templates/_helpers.tpl
```diff
@@ -167,42 +183,47 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
...
-{{/*
-  This template renders the URI for RabbitMQ's API endpoint.
-*/}}
-{{- define "invenio.rabbitmq.apiUri" -}}
-  {{- $username := (include "invenio.rabbitmq.username" .) -}}
-  {{- $password := (include "invenio.rabbitmq.password" .) -}}
-  {{- $port := (include "invenio.rabbitmq.managementPort" .) -}}
-  {{- $hostname := (include "invenio.rabbitmq.hostname" .) -}}
-  {{- printf "http://%s:%s@%s:%v/api/" $username $password $hostname $port }}
+{{- define "invenio.config.queue" -}}
```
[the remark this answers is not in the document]
lindhe: https://github.com/inveniosoftware/helm-invenio/pull/152#discussion_r2003222389

charts/invenio/values.yaml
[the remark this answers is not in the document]
egabancho: In general, I like explicit over implicit 😉. Also, I took inspiration from bitnami on this one too, i.e. https://github.com/bitnami/charts/blob/main/bitnami/airflow/values.yaml#L3134 

I think it might be important down the road when we autogenerate the documentation.

And! It does say `All of these values are only used when rabbitmq.enabled is set to false`

charts/invenio/templates/_helpers.tpl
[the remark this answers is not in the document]
egabancho: Fine by me, but shouldn't it be `postString` instead? (camelcase)

charts/invenio/templates/_helpers.tpl
egabancho: I think it makes sense ... https://github.com/inveniosoftware/invenio-celery/blob/master/invenio_celery/config.py#L17 😂

egabancho:
I don't see any of these variables in any of the containers I have (using my old branch with these changes). But, judging from the name, they might be some k8s thingy?

charts/invenio/templates/_helpers.tpl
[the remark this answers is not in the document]
egabancho: Should we do the same thing for RMQ's port?

charts/invenio/templates/_helpers.tpl
lindhe: Good argument! :D I'll see to it.

charts/invenio/values.yaml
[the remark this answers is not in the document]
lindhe: If the values are not used unless `rabbitmq.enabled=false`, then I think they should be commented out by default. I also prefer explicit, but this would explicitly set something that's not used, which I think is confusing. I'll comment them out, is  that OK?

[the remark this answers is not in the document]
lindhe: Another reason is that `externalOpensearch` is also commented out.

lindhe (approved):
If you are happy with the changes now, I think this is ready to merge. 🙂

max-moser:
i'm not familiar enough to make a call with helm charts, but it looks like there was a discussion and it got accepted – so i moved it to "to release"

lindhe:
@max-moser I still have some unresolved comments. No release yet.

charts/invenio/values.yaml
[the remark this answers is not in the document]
egabancho: I kind of disagree, but I don't have strong feelings about it 😉 

About Opensearch, it is not commented out, it's just empty! 😂 We are looking for a provider. As soon as I have something to work with, I'll "enhance" that section.

Now, to the issue at hand. I believe they are better there, explicitly, because when I want to use that block, I can now quite easily see which values I can set and how they look. If we remove that, then I'd have to start looking around the templates.
We already do that elsewhere. Take `invenio.sentry`. It is disabled by default, but all the values that can be set (with docstrings 😉) are there.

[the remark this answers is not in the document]
egabancho: Just to be clear, I am fine with the values being commented out, i.e., `# foo: bar`, not removing them.

egabancho:
I can't approve the PR because I made it 😛 
@lindhe, if you're happy with it, I am happy with it.

charts/invenio/values.yaml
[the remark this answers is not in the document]
lindhe: > About Opensearch, it is not commented out, it's just empty!

Yeah, I misspoke. I meant that the `externalOpensearch` is the empty object, but it has some suggested values included as comments next to it.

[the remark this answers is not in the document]
lindhe: > Now, to the issue at hand. I believe they are better there, explicitly, because when I want to use that block, I can now quite easily see which values I can set and how they look. If we remove that, then I'd have to start looking around the templates.

In my opinion, that's solved in a good way by providing comments. Easy to find and use, yet unmistakably non-active values!


>  We already do that elsewhere. Take invenio.sentry. It is disabled by default, but all the values that can be set (with docstrings 😉) are there.

Great point! I assumed Invenio required those values, so I didn't dare unset them despite not using Sentry. When I started using Invenio, I didn't know if sentry was https://sentry.io or some internal component in Invenio. So I got completely fooled by the default values being there. It's so hard to determine if a value in Helm is used or not…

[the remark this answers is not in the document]
lindhe: > I kind of disagree, but I don't have strong feelings about it 😉

Great. Because the more I think about it, the more I feel that I probably have strong-ish feelings about it. 😄 So I'll go ahead and comment them out.

lindhe:
> I can't approve the PR because I made it 😛 @lindhe, if you're happy with it, I am happy with it.

Thanks. I'll add the last change with comments that I mentioned, then I'll approve. Then you merge if you are happy with my changes. 😛

lindhe:
Totally missed your last two commits! I think this looks good now. I'll just one quick test to see that I didn't miss something obvious, then I'll approve.

lindhe (approved):

PR #171, opened 2025-03-14
add extraEnvVars to the invenio block and deprecate invenio.extra_env_from_secret

egabancho:


charts/invenio/templates/install-init-job.yaml
```diff
@@ -44,6 +44,13 @@ spec:
           value: {{ include "invenio.rabbitmq.uri" . }}
         - name: INVENIO_SQLALCHEMY_DATABASE_URI
           value: {{ include "invenio.sqlAlchemyDbUri" . }}
+        {{- range .Values.invenio.extraEnvFromSecret }}
+        - name: {{ tpl .name $ }}
+          valueFrom:
+            secretKeyRef:
+              name: {{ tpl .valueFrom.secretKeyRef.name $ }}
+              key: {{ tpl .valueFrom.secretKeyRef.key $ }}
+        {{- end }}
```
lindhe, line 53: In my opinion, it's cleaner to use a `with` and let the user define all elements instead of iterating over the list and reconstructing the element. This reduces the need for us to construct new objects and lets the user lookup how a `valueFrom` should look like. But I guess the variable should be renamed, if we change it so drastically.

lindhe: Sure. But I just want emphasize that while it sounds similar, they are subtly different.

I named my value `extraEnvFrom` because it inserts _extra_ elements to the `envFrom` list (also, this follows the naming pattern in Bitnami charts). But it cannot pick a specific key from a Secret and add that as an extra element to the `env` list.

Contrary, this value (`extra_env_from_secret`/`extraEnvFromSecret`) inserts extra elements to the `env` list _from_ some keys in some secrets. But the user has to specify every individual key and its corresponding `env` name, instead of including all of them as in `envFrom`.

egabancho: We are on the same page.
How would you like to proceed then?

lindhe: I suggest we mark `extra_env_from_secret` for deprecation, just like you have already done here, and introduce a new value `extraEnv` (or `extraEnvVars`, if we want to follow Bitnamis naming more closely) which is a list of objects that gets appended to the `env` field in each respective container.

Then this PR would no longer be a pure case change, but I don't think it's important to do as long as we take care to deprecate any old values we plan on removing.

What do you think?

egabancho: Funny enough, we have `extraEnvVars` on each of the services, i.e., https://github.com/inveniosoftware/helm-invenio/blob/master/charts/invenio/values.yaml#L205.
I'll add the general one and update the deprecation message.

charts/invenio/values.yaml
```diff
@@ -99,12 +99,11 @@ invenio:
   ## @param invenio.extraConfig Extra environment variables (templated) to be added to all the pods.
   ##
   extraConfig: {}
+  ## @param invenio.extra_env_from_secret DEPRECATED: invenio.extraEnvFromSecret instead
```
lindhe, line 102:
```suggestion
  ## @param invenio.extra_env_from_secret DEPRECATED: Use `invenio.extraEnvFrom` or `invenio.extraEnvVars` instead.
```

This this is a typo?

charts/invenio/values.yaml
```diff
@@ -99,12 +99,11 @@ invenio:
...
   extra_env_from_secret: []
-    # - name: NAME_OF_MY_ENV_VAR
-    #   valueFrom:
-    #     secretKeyRef:
-    #       name: name-of-my-secret
-    #       key: KEY_IN_MY_SECRET
+  ## @param invenio.extraEnvVars Extra environment variables (templated) to be added to all the pods.
```
lindhe, line 104: > templated

Are they? 🤔

egabancho: None of them are actually, not even the existing ones 😂 
Should I update the docs everywhere or add `tpl` instead? 🤔

lindhe: Good question. I lean towards just updating the comments/docs, because adding `tpl` adds complexity and I prefer doing that only when  there's a clear reason to do so. But I'm open to either alternative.

lindhe (approved):

PR #172, opened 2025-03-14
values: add nodeSelector and tolerations

egabancho:


lindhe:
I guess this should be for flower too?

lindhe (approved):
Please review my addition to the flower component. Other than that, I've just removed some trailing whitespace and rebased.

Issue #117, opened 2024-03-25
Secret handling - where should passwords, etc. be added


jaolwi:
I think the best way to handle this is to leave it up to the user whether he enters the password as plaintext in the values.yaml or integrates it via an existing secret. Here bitnami also has good examples. This procedure could also be useful for all backend service connection strings like postgresql.

Example:

```yaml
datacite:
    ## @param invenio.datacite.enabled Enable DataCite provider
    ##
    enabled: false
    ## @param invenio.datacite.username Datacite username
    ##
    username: ""
    ## @param invenio.datacite.password Datacite password 
    ##
    password: ""
    ## @param invenio.datacite.existingSecret Existing secret name for datacite username and password
    ##
    existingSecret: ""
    ## @param invenio.datacite.secretKeys.usernameKey Name of key in existing secret to use for username. Only used when `invenio.datacite.existingSecret` is set.
    ## @param invenio.datacite.secretKeys.passwordKey Name of key in existing secret to use for password. Only used when `invenio.datacite.existingSecret` is set.
    ##
    secretKeys:
      usernameKey: ""
      passwordKey: ""
```

lindhe:
Relates to https://github.com/inveniosoftware/helm-invenio/issues/112#issuecomment-2017524354

lindhe:
One area where I would like to see some improved secrets management is for providing custom environment variables to Invenio. Today, custom environment variables can be added via a ConfigMap `invenio-config`:

https://github.com/inveniosoftware/helm-invenio/blob/eb6544fdeb7e762972db1dae26f848c2f2e4c029/charts/invenio/templates/web-deployment.yaml#L28-L30

But there's no way to add custom values via a Secret instead.

I see no reason to keep the ConfigMap, I think that it can be replaced by a secret altogether (replacing `configMapRef` with `secretRef` in the example above). But if we want to keep it, we should at least add a Secret too!

jaolwi:
I absolutely agree with you. To make the whole thing even more flexible for the user, you could also implement it in this way:

values.yaml
```yaml
 ## @param invenio.envFrom Load environment variables from kubernetes secret or config map to each invenio pod (web. worker, worker-beat)
  ##
  envFrom: []
    # - secretRef:
    #     name: env-secret
    # - configMapRef:
    #     name: config-map
```

web-deployment.yaml.yaml
```yaml
command:...
envFrom:
  {{- with .Values.invenio.envFrom }}
[...]

PR #173, opened 2025-03-14
invenio: refactor datacite block

egabancho:
* Deprecates non-camelcase datacite variables.

* Addresses secret handling as described in #117.

### DIFF
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -334,7 +334,7 @@ INVENIO_SEARCH_HOSTS: {{ printf "[{'host': '%s'}]" (include "invenio.opensearch.
 INVENIO_SITE_HOSTNAME: '{{ include "invenio.hostname" $ }}'
 INVENIO_SITE_UI_URL: 'https://{{ include "invenio.hostname" $ }}'
 INVENIO_SITE_API_URL: 'https://{{ include "invenio.hostname" $ }}/api'
-INVENIO_DATACITE_ENABLED: "False"
+INVENIO_DATACITE_ENABLED: {{ternary "True" "False" .Values.invenio.datacite.enabled | quote }}
 INVENIO_LOGGING_CONSOLE_LEVEL: "WARNING"
 {{- end -}}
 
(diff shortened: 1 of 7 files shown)


egabancho:
I haven't tested this one yet.

charts/invenio/templates/_helpers.tpl
```diff
@@ -334,7 +334,7 @@ INVENIO_SEARCH_HOSTS: {{ printf "[{'host': '%s'}]" (include "invenio.opensearch.
 INVENIO_SITE_HOSTNAME: '{{ include "invenio.hostname" $ }}'
 INVENIO_SITE_UI_URL: 'https://{{ include "invenio.hostname" $ }}'
 INVENIO_SITE_API_URL: 'https://{{ include "invenio.hostname" $ }}/api'
-INVENIO_DATACITE_ENABLED: "False"
+INVENIO_DATACITE_ENABLED: "{{ .Values.invenio.datacite.enabled }} "
```
lindhe, line 337: Before it was `"False"`, now it's `"false"`. Are both valid?

lindhe (approved):
Just check my comment. Other than that, I think this looks good.

charts/invenio/templates/_helpers.tpl
```diff
@@ -334,7 +334,7 @@ INVENIO_SEARCH_HOSTS: {{ printf "[{'host': '%s'}]" (include "invenio.opensearch.
 INVENIO_SITE_HOSTNAME: '{{ include "invenio.hostname" $ }}'
 INVENIO_SITE_UI_URL: 'https://{{ include "invenio.hostname" $ }}'
 INVENIO_SITE_API_URL: 'https://{{ include "invenio.hostname" $ }}/api'
-INVENIO_DATACITE_ENABLED: "False"
+INVENIO_DATACITE_ENABLED: {{ternary "True" "False" .Values.invenio.datacite.enabled | quote }}
```
egabancho, line 337: We need this to be a string, hence the `quote`. Otherwise, it will get translated into `true` and the final value would be a string inside Invenio:
```
In [1]: app.config['DATACITE_ENABLED']
Out[1]: 'true'
```

charts/invenio/templates/_helpers.tpl
egabancho: They are not! The lower-case ones won't get interpreted as booleans, messing up the whole thing. 
I fixed that up already.

charts/invenio/templates/_helpers.tpl
lindhe, line 337:
```suggestion
INVENIO_DATACITE_ENABLED: {{ ternary "True" "False" .Values.invenio.datacite.enabled | quote }}
```

lindhe (approved):
I haven't had time to test this PR, but I think it looks good. Feel free to merge; we can fix bugs later if we discover them.

If I do get time to test run this, I'll write. But I approve it anyway, just check my last remaining suggestion.

PR #186, opened 2025-04-21
datacite: add configuration variables

egabancho:
It adds all the DataCite configuration variables to the `invenio.datacite` block, which, IMHO, is better than having them spread all around your `values.yaml`.


charts/invenio/values.yaml
```diff
@@ -105,6 +105,21 @@ invenio:
...
+    prefix: ""
+    ## @param invenio.datacite.testMode DataCite test mode enabled.
+    ## This has to be a string value, invenio will evaluate this as a python expresion an transform it into a boolean.
+    ## Example:
+    ## testMode: "False"
+    ##
+    testMode: ""
+    ## @param invenio.datacite.format A string used for formatting the DOI.
+    ##
+    format: ""
```
egabancho, line 119: I am not sure if we want to set defaults here, i.e. https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/config.py#L466, or just check in the chart if the value is set (this is what I did)

charts/invenio/values.yaml
```diff
@@ -105,6 +105,21 @@ invenio:
...
+    ## @param invenio.datacite.centerSymbol DataCite data center symbol.
+    ##
+    centerSymbol: ""
```
lindhe, line 122: Sorry for my ignorance, but this reads like BIOS tool tips for me. 😁

> datacite.centerSymbol DataCite data center symbol.

Can we improve the comment with some info about what a "DataCite data center symbol" is?

Also, right now we call it both "data center symbol" and just "centerSymbol". Let's consolidate that.

lindhe:
Left some minor comments. Once the conversations get resolved I'll approve this.

charts/invenio/values.yaml
egabancho: I believe it's a bit more than that. It actually formats the DOI. 
If you set it to `{prefix}/repo.{id}` the DOIs will look like `10.5281/repo.abc1-xyz9`.

The syntax is described on the link above, more or less 😅 You can see an example in action here https://github.com/zenodo/zenodo-rdm/blob/master/invenio.cfg#L327

charts/invenio/values.yaml
egabancho: 😂 
For all these variables, I just copied the first line of the config from Invenio-RDM-Records, in this case, https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/config.py#L481.

I looked into linking to docs but ... https://invenio-rdm-records.readthedocs.io/en/latest/configuration.html and didn't want to link to code directly because lines might change in the future.

lindhe: Ah, another case of the good ol' _undocumented parameters_ – nice!

My primary suggestion is that we copy or get inspired by the code you linked, and write our own comments based on that.

In second plance, I think a permalink to the code file is fine.

If you disagree with both of those, having no comments is not a deal breaker for me. At least that's easy to maintain. 😄

egabancho: I added a link to the general RDM documentation at the top (I didn't think about this one before) and also referenced the config variable that each key will set.
I think it's much better than before now 😉

lindhe: Big improvement. 👍

lindhe (approved):

PR #192, opened 2025-04-24
haproxy removal

egabancho:


lindhe (approved):
I'm so happy about this change! This will reduce complexity for future improvements, I believe.

LGTM!

PR #193, opened 2025-04-25
config: add TRUSTED_HOSTS for Flask v3 compatibility

egabancho:
I am not simply replacing `INVENIO_APP_ALLOWED_HOSTS` to keep compatibility with v12 while allowing v13 😉

### DIFF
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -393,6 +393,7 @@ Invenio basic configuration variables
 {{- define "invenio.configBase" -}}
 INVENIO_ACCOUNTS_SESSION_REDIS_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/1'
 INVENIO_APP_ALLOWED_HOSTS: '["{{ include "invenio.hostname" $ }}"]'
+INVENIO_TRUSTED_HOSTS: '["{{ include "invenio.hostname" $ }}"]'
 INVENIO_CACHE_REDIS_HOST: '{{ include "invenio.redis.hostname" . }}'
 INVENIO_CACHE_REDIS_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/0'
 INVENIO_CELERY_RESULT_BACKEND: 'redis://{{ include "invenio.redis.hostname" . }}:6379/2'


Samk13:
Thanks for that, but I'm testing by adding both values, and it seems that even when both are provided, the warnings still appear. See below: 👇
```py
TRUSTED_HOSTS = ['0.0.0.0', 'localhost', '127.0.0.1']
APP_ALLOWED_HOSTS = ['0.0.0.0', 'localhost', '127.0.0.1']
```
Screenshot 2025-04-25 at 18 03 58

So we are not benefiting from this much, one idea is to mark the legacy `APP_ALLOWED_HOSTS` as deprecated, encouraging users to migrate to `TRUSTED_HOSTS`. 
🤔Or are there neater ways to enforce the use of the new variable as replacing an old one but still allowing both @egabancho  @lindhe?

egabancho:
My idea was to make the chart compatible with v12 and v13 for a while.
I believe it doesn't hurt to have both variables, other than the warning messages.

Samk13:
My vote is to drop `APP_ALLOWED_HOSTS` altogether from the chart and rely exclusively on the new `TRUSTED_HOSTS`. That keeps the template clean. We can roll this into the next Helm-chart major bump together with the v13 upgrade and flag the removal clearly in the release notes/changelog so users have a heads-up.

lindhe:
I don't have a particular opinion on the environment variables, since I don't have much insight into their effects. But it sounds to me like we need a strategic decision on backwards compatibility w.r.t. older versions of Invenio.

Do I understand you correctly that `TRUSTED_HOSTS` was introduced in v13 and is not supported by v12? If so, I feel strongly that we should not drop `APP_ALLOWED_HOSTS` yet, because v12 is still in use. But again – this needs to be a strategic decision regarding which versions of Invenio we intend for this chart to support.

That decision may be pretty tricky to come up with, considering how common it seems to be for users to provide their own home brew build of Invenio. We do have a demo image as the default image right now, and clearly we should ensure that this works out-of-the-box. But it's hard to say for sure which variations and/or versions should be supported.

egabancho:
@ Samk13, the goal is to make the chart compatible with v12 and 13 for a while.
I know we'll get the warning on v13, but he current _official_ release is v12, and the chart should work nicely with it. 
Let's say that we are preparing ourselves for the future. Although we know the future is here, I'm running v13 already 😉 

WDYT? Did we convince you? 😇

lindhe (approved):

Samk13 (approved):

PR #195, opened 2025-04-25
nginx: rename extra_server_config

egabancho:


charts/invenio/templates/NOTES.txt
```diff
@@ -135,3 +135,11 @@ DEPRECATION WARNING:
     `image.pullSecrets` and it will be removed in a future release.
 
 {{- end }}
+
+{{- if .Values.nginx.extra_server_config }}
+
+DEPRECATION WARNING:
+    `nginx.extra_server_config` has been renamed in favor of
+    `nginx.extraEnvFrom` will be removed in a future release.
```
lindhe, line 143: I know these lines are coped from other places, but I have to admit they don't sit great with me. The last line makes it look like "`nginx.extraEnvFrom` will be removed in a future release" which is the opposite of what we want to communicate.

You may leave it like this if you want, just to conform with the rest of the messages. And maybe it's not worth changing since they will all be removed for 1.0 release. I just wanted to highlight this potential issue.

Feel free to mark this comment as resolved if you agree.

lindhe (approved):

PR #210, opened 2025-07-14
feat: new terminal deployment to run maintenance tasks

egabancho:


charts/invenio/templates/terminal-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+          readinessProbe: {{- toYaml .Values.kerberos.readinessProbe | nindent 12 }}
+          {{- end }}
+          {{- if .Values.kerberos.livenessProbe }}
+          livenessProbe: {{- toYaml .Values.kerberos.livenessProbe | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - name: kerberos-credentials-cache
+              mountPath: /tmp
+        {{- end }}
+
```
lindhe, line 131: Accidental blank newline?

```suggestion
```

charts/invenio/templates/terminal-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+        {{- if .Values.invenio.vocabularies }}
+        - name: vocabularies
+          configMap:
+            name: "{{ include "invenio.fullname" . }}-vocabularies"
+        {{- end }}
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+
```
lindhe, line 169: Remove trailing newline:

```suggestion
```

scripts/console.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
lindhe, line 1: Should we add ShellCheck to the CI?

scripts/console.sh
```diff
@@ -0,0 +1,67 @@
...
+  if [[ "$confirm" =~ ^[Yy]$ ]]; then
+    echo "⏬ Scaling down terminal deployment..."
+    kubectl scale deployment -n "$NAMESPACE" -l "$LABEL" --replicas=0
+  else
+    echo "✅ Pod left running."
+  fi
+else
+  echo "✅ Pod left running."
+fi
+
```
lindhe, line 67: Remove trailing newline:


```suggestion
```

charts/invenio/templates/terminal-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+          {{- if .Values.kerberos.readinessProbe }}
+          readinessProbe: {{- toYaml .Values.kerberos.readinessProbe | nindent 12 }}
+          {{- end }}
+          {{- if .Values.kerberos.livenessProbe }}
+          livenessProbe: {{- toYaml .Values.kerberos.livenessProbe | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - name: kerberos-credentials-cache
+              mountPath: /tmp
+        {{- end }}
```
lindhe, line 130: I'm unfamiliar with kerberos. Why is it needed here?

charts/invenio/templates/terminal-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+                name: {{ include "invenio.secretName" . }}
+            {{- with .Values.invenio.extraEnvFrom }}
+            {{- tpl (. | toYaml) $ | nindent 12 }}
+            {{- end }}
+            {{- with .Values.terminal.extraEnvFrom }}
+            {{- tpl (. | toYaml) $ | nindent 12 }}
+            {{- end }}
+          env:
+          - name: TZ
+            value: {{ required "Missing .Values.global.timezone" .Values.global.timezone }}
```
lindhe, line 68: I think it would be nicer to just have default values that sets this, rather than having it be "hard-coded". I don't feel strongly about it, WDYT?

lindhe (changes requested):
Just found some minor things to comment on. But I like the feature overall.

charts/invenio/values.yaml
```diff
@@ -559,6 +559,132 @@ workerBeat:
     ##
     strategy: null
 
+## @section Terminal
+## Pods for runnning maintenance tasks.
+terminal:
+  ## @param terminal.enabled Enable the terminal deployment.
+  ##
+  enabled: true
```
lindhe, line 567: Is this ever used?

Also: Does it make sense to enable by default? Seems like an unnecessary security risk and resource utilization that should be up to people to opt-in for.

charts/invenio/templates/terminal-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+        {{- tpl (toYaml .) $ | nindent 8 -}}
+      {{- end }}
+      {{- with .Values.terminal.tolerations }}
+      tolerations:
+        {{- tpl (toYaml .) $ | nindent 8 -}}
+      {{- end }}
+      containers:
+        - name: terminal
+          image: {{ default .Values.terminal.image (include "invenio.image" .) }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
```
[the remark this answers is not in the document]
egabancho: The image bit was a "carry over" from the other deployments. We have already discussed that we should use only one image for everything, https://github.com/inveniosoftware/helm-invenio/pull/174.

One could argue that the "terminal" one is a special one, and we might want a different image with developer-friendly tools installed. 
IMHO, I think we should leave it with the default image and see how folks are using it. If necessary, we will add it back in.

charts/invenio/templates/terminal-deployment.yaml
egabancho: BWT, I simply copied everything we have from worker/web.

scripts/console.sh
egabancho: This is an excellent point!
I already added the action.

PR #200, opened 2025-05-20
feat(Web): restart Pod when templated invenio-configmap.yaml is chang…

Syndlex:
…ing.

Pods do not restart when configmap is restarted this automated the process.

:heart: Thank you for your contribution!

### Description

> Please describe briefly your pull request.

feat(Web): restart Pod when templated invenio-configmap.yaml is changing.

Pods do not restart when configmap is restarted this automated the process.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

PR #211, opened 2025-07-14
pods: add custom labels and annotations

### LINKED ISSUE #115: Extension points for custom labels and annotations in each resource
To make the chart as customizable as possible for users, it would be good to create entrypoints in all resources for labels and annotations. Annotations and labels are often used by clusters to implement network rules or to inject sidecars.

Possible solution:

There are examples in the postgresql charts from Bitnami.
The best way to implement this is probably to use the "with" statement.

```yaml

#values.yaml
web:
 extraLabels: {}
 extraAnnotations: {}

#web-deploy.yaml
labels:
  {{- with .Values.web.extraLabels }}
    {{- toYaml . | nindent 8 }}
  {{- end }}
annotations:
  {{- with .Values.web.extraAnnotations }}
    {{- toYaml . | nindent 8 }}
  {{- end }}
```

egabancho:
closes #115 #200

### DIFF
--- a/charts/invenio/templates/web-deployment.yaml
+++ b/charts/invenio/templates/web-deployment.yaml
@@ -30,6 +30,9 @@ spec:
     metadata:
       annotations:
         app.kubernetes.io/component: web
+        checksum/invenio-config: {{ include (print $.Template.BasePath "/invenio-configmap.yaml") . | sha256sum }}
+        checksum/nginx-configmap: {{ include (print $.Template.BasePath "/nginx-configmap.yaml") . | sha256sum }}
+        checksum/uwsgi-configmap: {{ include (print $.Template.BasePath "/uwsgi-configmap.yaml") . | sha256sum }}
         {{- with .Values.web.podAnnotations }}
           {{- tpl (toYaml .) | nindent 8 }}
         {{- end }}
@@ -38,11 +41,6 @@ spec:
           {{- tpl (toYaml .) | nindent 8 }}
         {{- end }}
         app.kubernetes.io/component: web
-      annotations:
-        app.kubernetes.io/component: web
-        {{- with .Values.web.podAnnotations }}
-          {{- tpl (toYaml .) | nindent 8 }}
-        {{- end }}
     spec:
       terminationGracePeriodSeconds: {{ .Values.web.terminationGracePeriodSeconds }}
       {{- with .Values.web.nodeSelector }}
--- a/charts/invenio/templates/worker-beat-deployment.yaml
+++ b/charts/invenio/templates/worker-beat-deployment.yaml
@@ -33,6 +33,7 @@ spec:
           {{- tpl (toYaml .) | nindent 8 }}
         {{- end }}
         app.kubernetes.io/component: worker-beat
+        checksum/invenio-config: {{ include (print $.Template.BasePath "/invenio-configmap.yaml") . | sha256sum }}
       labels:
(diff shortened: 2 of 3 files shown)


charts/invenio/templates/worker-deployment.yaml
```diff
@@ -29,7 +29,14 @@ spec:
   template:
     metadata:
       labels:
+        {{- with .Values.worker.podLabels }}
+          {{- tpl (toYaml .) | nindent 10 }}
+        {{- end }}
         app.kubernetes.io/component: worker
+      {{- with .Values.worker.podAnnotations }}
+      annotations:
+          {{- tpl (toYaml .) | nindent 8 }}
```
lindhe, line 38: Is this the correct indentation?

charts/invenio/templates/worker-beat-deployment.yaml
```diff
@@ -29,7 +29,14 @@ spec:
   template:
     metadata:
       labels:
+        {{- with .Values.workerBeat.podLabels }}
+          {{- tpl (toYaml .) | nindent 10 }}
+        {{- end }}
         app.kubernetes.io/component: worker-beat
+      {{- with .Values.workerBeat.podAnnotations }}
+      annotations:
+          {{- tpl (toYaml .) | nindent 8 }}
```
lindhe, line 38: Indentation?

charts/invenio/templates/worker-beat-deployment.yaml
```diff
@@ -29,7 +29,14 @@ spec:
...
     metadata:
       labels:
+        {{- with .Values.workerBeat.podLabels }}
+          {{- tpl (toYaml .) | nindent 10 }}
+        {{- end }}
         app.kubernetes.io/component: worker-beat
+      {{- with .Values.workerBeat.podAnnotations }}
+      annotations:
+        {{- tpl (toYaml .) | nindent 8 }}
+      {{- end }}
```
lindhe, line 39: Why is `nindent` different for labels and annotations? 🤔 

Also, as a very minor nitpick: Can we have annotations first just so the fields are sorted alphabetically?

charts/invenio/templates/web-deployment.yaml
```diff
@@ -29,7 +29,17 @@ spec:
   template:
     metadata:
       labels:
+        {{- with .Values.web.podLabels }}
+          {{- tpl (toYaml .) | nindent 10 }}
+        {{- end }}
+        app.kubernetes.io/component: web
+      annotations:
+        checksum/nginx-configmap: {{ include (print $.Template.BasePath "/nginx-configmap.yaml") . | sha256sum }}
+        checksum/uwsgi-configmap: {{ include (print $.Template.BasePath "/uwsgi-configmap.yaml") . | sha256sum }}
```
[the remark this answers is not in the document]
lindhe: I agree, it seems to belong to that PR.

Also, more broadly regarding auto restarts upon config changes: I think we have to accept that there's no perfect solution. Personally, considering any solution would be imperfect, I prefer having simpler templates since we have to maintain them. But I don't feel very strongly about it, so if you prefer that we have some good enough mechanism for it then that's OK by me.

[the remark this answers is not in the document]
egabancho: I'll add this to #200 and rebase once this one gets merged.

lindhe (approved):

PR #236, opened 2026-03-05
fix: add tpl to extra volumes and remove range

egabancho:
Sorry about this one, apparently nobody tested the volume mount before (including me), and building on the existing configuration was not the best idea 😅


lindhe (approved):
Sorry for not catching that. 🙃 Your changes here look reasonable, but I'm crammed this week so I'll leave it to you to test it out for real before merging.

egabancho:
It's running on my test instance this time 😉. I'll make a patch release.

Issue #118, opened 2024-03-25
Connection configuration for external and subchart backend services (postgreqsl, opensearch, s3, etc)


jaolwi:
The configuration for each backend service should be available under `invenio.` and not in the root of values.yaml. We can use functions to dynamically assemble the connection strings and also connect external databases, etc. With `enable: false`-like parameter we could then ignore individual backend services. Or if the subchart of the backend service is in use, the chart ignores this section completly.

Example:

```yaml
invenio:
  rabbitmq:
      ## @param invenio.rabbitmq.host Rabbitmq hostname. Only used when `global.rabbitmqEnabled` is false.
      ##
      host: ""
      ## @param invenio.rabbitmq.port Rabbitmq port. Only used when `global.rabbitmqEnabled` is false.
      ##
      port: ""
      ## @param invenio.rabbitmq.username Rabbitmq username. Only used when `global.rabbitmqEnabled` is false.
      ##
      username: ""
      ## @param invenio.rabbitmq.password Rabbitmq password. Only used when `global.rabbitmqEnabled` is false.
      ##
      password: ""
      ## @param invenio.rabbitmq.existingSecret Existing secret name for rabbitmq password. Only used when `global.rabbitmqEnabled` is false.
      ##
      existingSecret: ""
      ## @param invenio.rabbitmq.secretKeys.passwordKey Name of key in existing secret to use for password. Only used when `invenio.rabbitmq.existingSecret` is set and `global.rabbitmqEnabled` is false.
      ##
      secretKeys:
        passwordKey: ""

```

egabancho:
Coming from #136 and a question that @lindhe [asked](https://github.com/inveniosoftware/helm-invenio/issues/136#issuecomment-2517118424).

A `global` key might make sense but for a slightly different purpose. I'm not sure if we actually need one right now, though. 
I don't think it should replace the `invenio` one. This is what I believe should be the structure that we should try to go to:

```
## @section Global parameters
## Global Docker image parameters (inspire by bitnami charts)
global:
   ...

## @section Invenio parameters

image:
  registry: 
  repository:
  ...

## Common invenio app parameters
invenio:
   ...

# One section for each kind
web:
worker:
workerBeat:
```

I am on the fence about the init/install job. I'd love for it to be configurable, i.e. change the commands it runs. 
Probably, it requires its own section (more than just `true/false`). I'm not sure if it should be inside `invenio` or at the root level.

WDYT?

lindhe:
[...]

PR #244, opened 2026-05-28
feat: Enable more granual configuration of `opensearchExternal`

lindhe:
### Description

I recently configured our Invenio instance to use an externally deployed OpenSearch cluster. I found that the required configuration was a bit fiddly, so I think others would find it useful if it was better integrated in the Helm chart.

This change includes:

- Fixing a typo: `externalOpensearch` --> `opensearchExternal`
- Introducing some new values to let users configure access to authenticated OpenSearch clusters with TLS and custom port number.
- Dynamically building the `INVENIO_SEARCH_HOSTS` string.
- Mounting CA cert.
- Flexible methods to set credentials in plain text or via secrets.

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).

**Reminder**

### DIFF
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -264,7 +264,10 @@ Return the proper Invenio image name
   {{- if not .Values.opensearch.enabled }}
     {{- with $.Values.opensearchExternal -}}
       {{- if .auth.enabled -}}
-        {{- $params = append $params "'http_auth': ('$(OPENSEARCH_USERNAME)', '$(OPENSEARCH_PASSWORD)')" -}}
+        {{- $username := .auth.usernameEnv.name -}}
+        {{- $password := .auth.passwordEnv.name -}}
+        {{- $credentials := printf "'http_auth': ('$(%s)','$(%s)')" $username $password -}}
+        {{- $params = append $params $credentials -}}
       {{- end -}}
       {{- if .encryption.enabled -}}
         {{- $params = append $params "'use_ssl': True" -}}
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -286,15 +286,15 @@ Return the proper Invenio image name
   This template renders the extra environment variables for opensearchExternal
 */}}
 {{- define "invenio.opensearch.env" -}}
-{{- if not .Values.opensearch.enabled -}}
-{{- with $.Values.opensearchExternal -}}
-{{- if (dig "auth" "enabled" false .) -}}
-{{- $u := required "Missing .Values.opensearchExternal.auth.usernameEnv" .auth.usernameEnv -}}
-{{- $p := required "Missing .Values.opensearchExternal.auth.passwordEnv" .auth.passwordEnv -}}
-{{- list $u $p | toYaml -}}
-{{- end -}}
-{{- end -}}
-{{- end -}}
+  {{- if not .Values.opensearch.enabled -}}
+    {{- with $.Values.opensearchExternal -}}
+      {{- if (dig "auth" "enabled" false .) -}}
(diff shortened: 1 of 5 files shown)


egabancho (approved):
I haven't tested it, but it looks good.
I think this is the last piece of https://github.com/inveniosoftware/helm-invenio/issues/118 👏

charts/invenio/templates/_helpers.tpl
```diff
@@ -255,6 +255,81 @@ Return the proper Invenio image name
...
+        {{- $password := .auth.passwordEnv.name -}}
+        {{- $credentials := printf "'http_auth': ('$(%s)','$(%s)')" $username $password -}}
+        {{- $params = append $params $credentials -}}
+      {{- end -}}
+      {{- if .encryption.enabled -}}
+        {{- $params = append $params "'use_ssl': True" -}}
+        {{- $params = append $params (printf "'ca_certs': '%s'" .encryption.caCert.mountPath) -}}
+      {{- end -}}
+      {{- if .port -}}
+        {{- $params = append $params (printf "'port': '%d'" .port) -}}
```
lindhe: I feel compelled to double-check that this number gets treated right...

charts/invenio/templates/worker-deployment.yaml
```diff
@@ -189,6 +190,7 @@ spec:
           configMap:
             name: "{{ include "invenio.fullname" . }}-vocabularies"
         {{- end }}
+        {{- include "invenio.opensearch.volumes" . | nindent 8 }}
         {{- with .Values.invenio.extraVolumes }}
```
Samk13, line 195: Question: should we also include `invenio.opensearch.env` here similar to the web and worker-beat or is this intentional?

lindhe: I totally missed that one. Thank you very much. :+1:

lindhe: @egabancho There should be no need for OpenSearch in Flower, though. Right?

egabancho: correct, no OpenSearch inside the flower pod.

charts/invenio/templates/_helpers.tpl
lindhe: Caught it! :fist_raised: 

```diff
- ''port'': ''%!d(float64=9200)''
+ ''port'': ''9200''
```

charts/invenio/templates/_helpers.tpl
```diff
@@ -255,6 +255,81 @@ Return the proper Invenio image name
...
 
+#########################     SEARCH_HOSTS string     #########################
+{{/*
+  This template renders the string value for the [INVENIO_]SEARCH_HOSTS environment variable.
+*/}}
+{{- define "invenio.searchHostsString" -}}
+  {{- $params := list (printf "'host': '%s'" (include "invenio.opensearch.hostname" .)) -}}
+  {{- if not .Values.opensearch.enabled }}
+    {{- with $.Values.opensearchExternal -}}
+      {{- if .auth.enabled -}}
```
lindhe: Since `opensearchExternal.auth = nil` by default, this is likely to fail in an ugly way. I intend to catch this error better.

charts/invenio/templates/_helpers.tpl
lindhe, line 1: I've introduced some trailing whitespace and blank newlines with this change. I'll try to clean that up before merging.

lindhe: I think I was wrong (kind of; hard to rely details I'm tired) 😅

lindhe:
FYI: There's a bug with the environment variable string that I didn't catch. I was so occupied with reviewing the rendered output so I forgot to run it live before merging. :facepalm: 

Fix incoming.

PR #245, opened 2026-06-08
feat: add Invenio jobs scheduler deployment & workerBeat.commandOverride

mb-wali:
### Description

Adds a new `workerBeat.commandOverride` Helm value that allows overriding the default Celery Beat startup command.

This provides a simple escape hatch for deployments that need to customize Beat behavior (e.g. using a custom scheduler implementation) without modifying the chart templates.

In addition, this change introduces a dedicated **worker scheduler Deployment** to support running Invenio Jobs’ `RunScheduler` as a separate Celery Beat process. This enables clearer separation between standard Celery Beat usage and Invenio’s job scheduling system.

## Changes

- Added `workerBeat.commandOverride` to `values.yaml`
- Updated the worker-beat deployment template to use the override when provided
- Added documentation and an example configuration

* Added workerBeat.commandOverride to values.yaml
* Added workerScheduler configuration block to values.yaml
* Added new worker-scheduler Deployment template
* Updated scheduler Deployment to support:
  * workerScheduler.enabled
  * workerScheduler.class
  * workerScheduler.commandOverride
* Updated worker-beat deployment template to use override when provided

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**


egabancho:
I am wondering whether we should make the new scheduler the default one, the one that includes Invenio-Jobs' scheduler. After all, InvenioRDM instances will come with it "built in".

@lindhe, @utnapischtim, WDYT?

lindhe:
I don't have an opinion. I don't have enough insight into the application design of Invenio to really tell what would make more sense. I trust your judgement, though.

egabancho:
Hey @mb-wali, I had a chat with @utnapischtim about how the regular celery beat and the Invenio-Jobs scheduler relate, and it turns out I was wrong.
I thought the Job scheduler would call `super` at some point, thus replacing the default velery beat. It doesn't.

This basically means we will need to have different deployments, the one we already have for the regular celery beat, and a new one for the Job scheduler. This means adding a new section to the `values.yaml`, `scheduler:`?, similar to the beat one. Maybe there is room for some refactoring there? I'll let you decide depending on your time, etc. 😉 

Allowing custom commands for both cases might be a good idea. You've already done it for the beat, so let's keep it there.

charts/invenio/values.yaml
```diff
@@ -833,6 +833,165 @@ workerBeat:
...
+  ##
+  ## Example:
+  ## workerBeat:
+  ##   commandOverride: "celery -A invenio_app.celery beat --scheduler invenio_jobs.services.scheduler:RunScheduler --loglevel=INFO"
+  commandOverride: ""
+
+workerScheduler:
+  ## @param workerScheduler.enabled Enable the Invenio jobs scheduler.
+  ##
+  enabled: false
```
egabancho: I would make it enabled by default. Invenio-Jobs might not work otherwise.

egabancho:
@mb-wali, can you update the README using https://github.com/bitnami/readme-generator-for-helm ?

I typically do `node /path/to/readme-generator-for-helm/bin/index.js -v charts/invenio/values.yaml -r README.md`, it seems to be the fastest/simplest 😇

charts/invenio/templates/worker-scheduler-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+      imagePullSecrets:
+      - name: {{ .Values.worker.imagePullSecret }}
+      {{- end }}
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+  strategy:
+    type: Recreate # For data safety, beat must never have more than 1 replica!
+{{- end }}
```
lindhe: Seems to be a missing newline at the end.

https://gist.github.com/olivertappin/19a95236d8b69f1c9eadfc799a9ab7fc

charts/invenio/values.yaml
```diff
@@ -833,6 +833,165 @@ workerBeat:
...
+  ## @param workerScheduler.livenessProbe.exec.command TODO:
+  ## @param workerScheduler.livenessProbe.initialDelaySeconds TODO:
+  ## @param workerScheduler.livenessProbe.timeoutSeconds TODO:
+  ##
+  livenessProbe:
+    exec:
+      command:
+        - /bin/bash
+        - -c
+        - "celery -A {{ .Values.worker.app }} inspect ping"
```
lindhe, line 957: I'm guessing this is correct, since `workerBeat` also uses the `worker.app` value. But I wanted to check with you guys, because I don't know the `celery` tool well enough to understand what it does. Perhaps it needs to align with the value used in the command of this container?

charts/invenio/values.yaml
```diff
@@ -833,6 +833,165 @@ workerBeat:
...
+    initialDelaySeconds: 20
+    timeoutSeconds: 30
+  ## @param workerScheduler.readinessProbe templated `readinessProbe` for the worker-beat container
+  ##
+  readinessProbe: null
+  ## @param workerScheduler.startupProbe templated `startupProbe` for the worker-beat container
+  ##
+  startupProbe: null
+
+  ## @extra workerScheduler.deploymentSpec DEPRECATED: Use `.Values.workerScheduler.deployment.spec` instead!
```
lindhe: Let's not add deprecated values to the new component! 😄 Same comment goes for other values too, just grep for it in your changes.

charts/invenio/templates/worker-scheduler-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+      {{- if .Values.persistence.enabled }}
+      - name: shared-volume
+        persistentVolumeClaim:
+          claimName: {{ .Values.persistence.name }}
+      {{- end }}
+      {{- if .Values.invenio.vocabularies }}
+      - name: vocabularies
+        configMap:
+          name: "{{ include "invenio.fullname" . }}-vocabularies"
+      {{- end }}
```
lindhe: Sorry if I'm asking silly questions here, but does the worker-scheduler need the shared volume and vocabularies volumes?

charts/invenio/templates/worker-scheduler-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+      {{- with .Values.worker.extraVolumes }}
+      {{- tpl (toYaml .) $ | nindent 6 }}
+      {{- end }}
+      {{- if .Values.worker.imagePullSecret }}
+      imagePullSecrets:
+      - name: {{ .Values.worker.imagePullSecret }}
+      {{- end }}
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
```
lindhe: Should probably not be `.Values.worker.*` values here.

charts/invenio/templates/worker-scheduler-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+  strategy:
+    type: Recreate # For data safety, beat must never have more than 1 replica!
```
lindhe: Is this true for `worker-scheduler`?

charts/invenio/templates/worker-scheduler-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+          "celery -A {{ .Values.worker.app }} beat --scheduler {{ .Values.workerScheduler.class }} -l {{ .Values.worker.log_level }}"
+        ]
+        {{- end }}
+        securityContext:
+          {{- with .Values.workerScheduler.securityContext }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
+        volumeMounts:
+        - name: var-run-celery
+          mountPath: {{ .Values.worker.run_mount_path }}
```
lindhe: I'd prefer if the mount path was specific to this component instead of sharing the worker's config. I think users will expect `.Values.worker` to change the worker pod, not other pods. Although, there are other places where shared values makes more sense (like in `command`). I'd like your input here, @egabancho.

lindhe (changes requested):

charts/invenio/templates/worker-scheduler-deployment.yaml
mb-wali: As this is a Celery Beat scheduler pod, running multiple Beat instances simultaneously may result in duplicate scheduling, causing periodic tasks to be enqueued more than once.

charts/invenio/templates/worker-scheduler-deployment.yaml
mb-wali: Not sure — @egabancho  do you have any context on this? I noticed the Beat pod also includes this configuration.

charts/invenio/templates/worker-scheduler-deployment.yaml
lindhe: Sounds reasonable.

egabancho: In line 35 there is already a `strategy`  key that is filled if `.Values.workerScheduler.deployment.spec.strategy` is set. I am wondering if we should remove line 35 or set `Recreate` as the default inside `values.yaml`.

charts/invenio/templates/worker-scheduler-deployment.yaml
egabancho: Theoretically, the schedulers don't need any volume to run, not even the persistence one.
The s[cheduler container inside the cookiecutter](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/docker-compose.full.yml#L133) doesn't mount any of the volumes.
I'd say we remove them from here and open an issue to clean up the worker beat , unless @mb-wali wants to do it here 😇

charts/invenio/values.yaml
```diff
@@ -833,6 +833,146 @@ workerBeat:
...
+  ## @param workerScheduler.enabled Enable the Invenio jobs scheduler.
+  ##
+  enabled: true
+  ## @param workerScheduler.class Scheduler class used by Celery beat.
+  ##
+  class: invenio_jobs.services.scheduler:RunScheduler
+  ## @param workerScheduler.extraEnvVars Extra environment variables to be added to the pods.
+  ##
+  extraEnvVars: []
+  ## @param workerScheduler.resources `resources` for the worker-beat container
```
egabancho: Can you update `worker-beat` to `scheduler` or something better? (same in other places)

charts/invenio/templates/worker-scheduler-deployment.yaml
egabancho: Practically, it's the Celery runtime directory, and there's no real reason to configure it differently per component. Typically, it's always `/var/run/celery` or similar.
I'd lean toward keeping the shared references for these low-level Celery settings (`app`, `log_level`, `run_mount_path`) since they're inherently tied to the same Celery application

charts/invenio/values.yaml
egabancho: It does need to align, and it already does (line 73 of the deployment).
`worker.app` is the Celery application entrypoint (i.e., `invenio_app.celery`) and it's the same Python application regardless of which Celery process runs it.

charts/invenio/templates/worker-scheduler-deployment.yaml
lindhe: Alright. Let's keep this as it is, until we have a clear reason to make it configurable in the future. :+1:

charts/invenio/templates/worker-scheduler-deployment.yaml
lindhe: Great idea. Doing that in a separate PR sounds like a good plan, reducing the scope of this one.

mb-wali: issue https://github.com/inveniosoftware/helm-invenio/issues/252

egabancho:
I added #253 as a follow-up to this one, together with #252.

egabancho (approved):

charts/invenio/values.yaml
```diff
@@ -861,6 +861,147 @@ workerBeat:
...
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+        - ALL
+  ## @extra workerScheduler.podSecurityContext securityContext for the scheduler pod
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
+  ## @param workerScheduler.podSecurityContext.runAsNonRoot TODO:
+  ## @param workerScheduler.podSecurityContext.runAsUser TODO:
+  ## @param workerScheduler.podSecurityContext.runAsGroup TODO:
+  ## @param workerScheduler.podSecurityContext.seccompProfile.type TODO:
```
lindhe: I don't want us to merge TODOs. Since these are standard Kubernetes fields, I don't think we need to provide documentation for them here, so I suggest we remove them (not `workerScheduler.podSecurityContext`, of course. Just the fields under it).

```suggestion
```

charts/invenio/values.yaml
```diff
@@ -861,6 +861,147 @@ workerBeat:
...
+  nodeSelector: {}
+  ## @param workerScheduler.tolerations Tolerations for workerScheduler pods assignment
+  ## Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+  ##
+  tolerations: []
+  ## @extra workerScheduler.livenessProbe templated `livenessProbe` for the worker-scheduler container
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+  ## @param workerScheduler.livenessProbe.exec.command TODO:
+  ## @param workerScheduler.livenessProbe.initialDelaySeconds TODO:
+  ## @param workerScheduler.livenessProbe.timeoutSeconds TODO:
```
lindhe: Same comment here; no need to explain them so better to remove these docs strings than introducing TODOs.


```suggestion
```

lindhe (changes requested):
Found some minor things to improve. And remember to render the README again after fixing them. I think the rest looks great, we can merge this after the fixes.

charts/invenio/values.yaml
egabancho: I removed the `TODO` and added `@skip` on the values we don't want to document (because they are the standard Kubernetes fields described on the link).
If I don't add the `@skip`, the readme generator complains about missing values.

Issue #22, opened 2021-12-08
ci for tests


ppanero:
- [ ] helm chart tests?
- [ ] helm lint

PR #247, opened 2026-06-24
add helm template testing and kubeconform validation

egabancho:
Replaces #237 Address #22

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@v4.2.0
         with:
-          version: v3.17.0
+          version: v3.21.2
 
       - uses: actions/setup-python@v5.3.0
         with:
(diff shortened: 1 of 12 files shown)


.github/workflows/lint-test.yml
```diff
@@ -33,6 +33,32 @@ jobs:
       - name: Run chart-testing (lint)
         run: ct lint --all
 
+      - name: Install Helm plugins
+        run: |
+          helm plugin install https://github.com/helm-unittest/helm-unittest.git
+          pip install pyyaml
+          helm plugin install https://github.com/jtyr/kubeconform-helm.git
```
lindhe: Just want to flag that jtyr/kubeconform-helm has pretty few stars and I've not heard of it before. Do we want to afflict ourselves with some rules for what dependencies we can rely on, or do we want to implicitly trust any reasonable looking project?

I don't mean this as a critique, I'm just thinking aloud here whether or not we need to set some ground rules to reduce the risk of supply chain attacks. That's hard to tell because I'm not exactly sure what the worst-case impact of a repo take-over would be.

lindhe (changes requested):

lindhe (changes requested):
I really like the initiative and I think most looks good. One comment to discuss. Just one request for change, which is actually my main hold-off before approval:

I would appreciate some docs on how to run this locally, either using the Helm plugins or perhaps using [ACT](https://github.com/nektos/act).

.github/workflows/lint-test.yml
egabancho: The actual tool has 3.1k: https://github.com/yannh/kubeconform. For a k8s tool that basically does linting, I'd say it is quite good 😅.

If you are talking about the helm plugin itself, then yes. It's relatively new, but it seems endorsed by kubeconform (link in the readme). 

I used the plugin for simplicity. I would have to install Kubeconform and run it with the helm output otherwise.

But, if you are not convinced, I understand your concerns and am happy to make the change.

lindhe: I'm convinced enough and I don't think this should be a blocker for this PR. I mostly wanted to bring it up because I think it's easily overlooked and I've grown a bit weary about GitHub Actions lately... :face_with_diagonal_mouth:

egabancho:
> I would appreciate some docs on how to run this locally, either using the Helm plugins or perhaps using [ACT](https://github.com/nektos/act).

I added a section to the README to describe how to set up your local dev environment to run the tests. I've never used ACT before, so I described how to run the same commands but locally. 
Let me know if that works?

---

I reverted to using `kuberconform` directly instead of the Helm plugin. The plugin requires installing `pyyaml` and, being a Python package, it opens the door to all sorts of things that we don't need on a Helm chart (i.e., virtual environments)

lindhe (approved):
Masterful. Well done.

PR #248, opened 2026-06-24
feat: add affinity configuration for web and worker pods

egabancho:
Allows users to configure pod affinity/anti-affinity rules for web, and worker deployments. This enables spreading replicas across nodes to improve availability

---

I based this branch on #247 to start adding tests right away.

### DIFF
--- a/charts/invenio/ci/full-values.yaml
+++ b/charts/invenio/ci/full-values.yaml
@@ -61,6 +61,18 @@ web:
     scaler_cpu_utilization: 65
     max_web_replicas: 10
     min_web_replicas: 2
+  affinity:
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: app.kubernetes.io/component
+                  operator: In
+                  values:
+                    - web
+            topologyKey: kubernetes.io/hostname
   nodeSelector:
     node-role: web
   tolerations:
@@ -85,6 +97,18 @@ web:
           maxUnavailable: 0
 
 worker:
+  affinity:
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: app.kubernetes.io/component
+                  operator: In
+                  values:
+                    - worker
+            topologyKey: kubernetes.io/hostname
   nodeSelector:
     node-role: worker
   tolerations:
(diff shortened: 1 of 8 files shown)


lindhe:
Excellent initiative, thanks for catching this! Kind of embarrassing that I hadn't thought of that before, this is pretty important for availability. I think we should set a `preferred` pod anti affinity by default, actually.

egabancho:
@lindhe, the affinity in the example would be ok as default?

I've seen [other charts simplifying this](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L246-L258) by using things like this:
```
# By default this will make sure two pods don't end up on the same node
# Changing this to a region would allow you to spread pods across regions
antiAffinityTopologyKey: "kubernetes.io/hostname"

# Hard means that by default pods will only be scheduled if there are enough nodes for them
# and that they will never end up on the same node. Setting this to soft will do this "best effort".
# Setting this to custom will use what is passed into customAntiAffinity.
antiAffinity: "soft"

# Allows passing in custom anti-affinity settings as defined in
# https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#types-of-inter-pod-affinity-and-anti-affinity
# Using this parameter requires setting antiAffinity to custom.
customAntiAffinity: {}
```
Do you think it'll make things cleaner or overcomplicate the chart logic?

lindhe:
> the affinity in the example would be ok as default?

Yes, I think those examples look great as defaults.

> I've seen other charts simplifying this […]
> 
> Do you think it'll make things cleaner or overcomplicate the chart logic?

I much prefer to let users provide the true Kubernetes field directly instead of abstracting it away.

- I really like that it provides more degrees of freedom to the end-user, without us having to implement new features in the Helm chart just because we didn't consider some edge-case that they need.
- As an end-user of a Helm chart, I really appreciate being able to draw on my knowledge about the Kubernetes API spec rather than having to learn a new API specification unique to a particular chart.

Creating abstractions can make sense. For example when a single parameter should make 10 different changes in different places. Or when something is critical for safe operation of the application, it's usually good to restrict the freedom of the end-user. I don't think that's the case here.

Glad that you asked, BTW. It's been on my mind lately, it felt good writing it down.

egabancho:
@lindhe, I've been thinking about what you said about the defaults. I'd be cautious about making the preferredDuringSchedulingIgnoredDuringExecution anti-affinity a default. Two main reasons:

- It might break single-node clusters like dev environments, Minikube, Kind, and small test clusters. Pods will schedule fine (it's `preferred`, not `required`), but the scheduler still pays the overhead of evaluating rules that can never be satisfied. More importantly, it sets an opinionated default that may surprise users who haven't thought about affinity.
- The README already says it's not production-ready, and I don't think it should state otherwise. Users who care about HA will configure affinity for their topology.

I'd keep `affinity: {}` as the default and leave the example as a comment for reference.

lindhe (approved):
I disagree, but it's no blocker.

The scheduling overhead for evaluating affinity for three pods is extremely small, I can't see this breaking stuff for anyone. But we can keep it as it is, that's fine. I agree that we don't need to make it prod ready, for the reason you mentioned.

Issue #249, opened 2026-06-24
Config: Add Redis authentication support.


plesoun-stein:
## Is your feature request related t/o a problem? Please describe.
The chart currently implements unauthenticated access to Redis by default.
For security compliance, Redis authentication is recommended.

## Describe the solution you'd like
The implementation follows the same pattern as the RabbitMQ configuration.

1. Create a set of helpers that render the authenticated Redis hostname.
2. Remove the Redis configuration from the `invenio.configBase` helper.
3. Add the `invenio.config.cache` helper to the deployments.

PR #251, opened 2026-06-26
feat: Add Redis authentication support

plesoun-stein:
inveniosoftware/helm-invenio#249
### Description

    Create a set of helpers that render the authenticated Redis hostname.
    Remove the Redis configuration from the invenio.configBase helper.
    Add the invenio.config.cache helper to the deployments.

### Checklist

- [x] I've added relevant test cases.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/charts/invenio/templates/_helpers.tpl
+++ b/charts/invenio/templates/_helpers.tpl
@@ -83,9 +83,88 @@ Return the proper Invenio image name
   {{- if .Values.redis.enabled }}
     {{- printf "%s-master" (include "common.names.fullname" .Subcharts.redis) }}
   {{- else }}
-      {{- required "Missing .Values.redisExternal.hostname" .Values.redisExternal.hostname }}
+    {{- required "\n\nYou have configured `.Values.redisExternal`, but  .Values.redisExternal.hostname is missing" .Values.redisExternal.hostname }}
   {{- end }}
 {{- end -}}
+
+{{- define "invenio.redis.password" }}
+  {{- if .Values.redis.enabled }}
+    {{- if not .Values.redis.auth.enabled }}
+      {{ print "value: \"\"" }}
+    {{- else if and .Values.redis.auth.enabled  .Values.redis.auth.password }}
+      {{ printf "value: %q" .Values.redis.auth.password }}
+    {{- else if and .Values.redis.auth.enabled .Values.redis.auth.existingSecret .Values.redis.auth.existingSecretPasswordKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.redis.auth.existingSecret }}
+      key: {{ .Values.redis.auth.existingSecretPasswordKey }}
+    {{- else }}
+      {{ fail "\n\n.Values.redis.auth.enabled is set to `True`, but the authentication configuration is missing.\nSpecify either .Values.redis.auth.password or both .Values.redis.auth.existingSecret and .Values.redis.auth.existingSecretPasswordKey." }}
+    {{- end }}
+  {{- else if .Values.redisExternal }}
+    {{- if .Values.redisExternal.password }}
+      {{ printf "value: %q" .Values.redisExternal.password }}
+    {{- else if and  .Values.redisExternal.existingSecret .Values.redisExternal.existingSecretPasswordKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.redisExternal.existingSecret }}
+      key: {{ .Values.redisExternal.existingSecretPasswordKey }}
+    {{- else }}
+      {{- fail "\n\nYou have configured `.Values.redisExternal`, but no authentication credentials were provided.\nSpecify either `.Values.redisExternal.password` or both `.Values.redisExternal.existingSecret` and `.Values.redisExternal.existingSecretPasswordKey`." }}
+    {{- end }}
+  {{- else }}
+    {{ fail "\nMissing Redis configuration.\nEnable Redis by setting `.Values.redis.enabled` to `true`, or configure `.Values.redisExternal`." }}
+  {{- end }}   
+{{- end }}
+
+{{/*
+  This template renders the port number for Redis.
+*/}}
+{{- define "invenio.redis.port" -}}
+    {{- print "6379" | quote -}}
+{{- end -}}
+
+{{/*
+  This template renders the protocol for Redis
+*/}}
+{{- define "invenio.redis.protocol" -}}
+  {{- print "redis" -}}
+{{- end -}}
+
+{{/*
+  Redis connection env section.
+*/}}
+{{- define "invenio.config.cache" -}}
+{{- if and (not .Values.redis.enabled) (not .Values.redisExternal) }}
+  {{ fail "\n\nMissing Redis configuration.\nEnable Redis by setting `.Values.redis.enabled` to `true`, or configure `.Values.redisExternal`." }}
+{{- end }}
+{{- $uri := "$(INVENIO_CONFIG_REDIS_PROTOCOL)://:$(INVENIO_CONFIG_REDIS_PASSWORD)@$(INVENIO_CONFIG_REDIS_HOST):$(INVENIO_CONFIG_REDIS_PORT)" -}}
+{{- $hostString := ":$(INVENIO_CONFIG_REDIS_PASSWORD)@$(INVENIO_CONFIG_REDIS_HOST)" -}}
+- name: INVENIO_CONFIG_REDIS_HOST
+  value: {{ include "invenio.redis.hostname" . }}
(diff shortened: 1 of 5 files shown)


plesoun-stein:
I changed `invenio.config.cache` to `invenio.config.redis` for clarity.

lindhe:
> [*] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/community/code-of-conduct/).

Please replace `*` with an `x` for each checkbox where you intend to confirm, so it renders correctly in GitHub.

Personally I don't care about this, but I want to avoid any confusion in case it's important for legal reasons. Not sure what the requirements are from the org's side, but it's easier to fix than looking up if it matters. :upside_down_face:

plesoun-stein:
> > [*] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/community/code-of-conduct/).
> 
> Please replace `*` with an `x` for each checkbox where you intend to confirm, so it renders correctly in GitHub.
> 
> Personally I don't care about this, but I want to avoid any confusion in case it's important for legal reasons. Not sure what the requirements are from the org's side, but it's easier to fix than looking up if it matters. 🙃

I left the Frontend, Translation, and License items unchecked because I believe they don't apply to this change.
I don't know whether I need to add tests, as I tested the changes in a live environment.

charts/invenio/templates/_helpers.tpl
```diff
@@ -83,10 +83,89 @@ Return the proper Invenio image name
...
+  {{- else if .Values.redisExternal }}
+    {{- if .Values.redisExternal.password }}
+      {{ printf "value: %q" .Values.redisExternal.password }}
+    {{- else if and  .Values.redisExternal.existingSecret .Values.redisExternal.existingSecretPasswordKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.redisExternal.existingSecret }}
+      key: {{ .Values.redisExternal.existingSecretPasswordKey }}
+    {{- else }}
+      {{- fail "\n\nYou have configured `.Values.redisExternal`, but no authentication credentials were provided.\nSpecify either `.Values.redisExternal.password` or both `.Values.redisExternal.existingSecret` and `.Values.redisExternal.existingSecretPasswordKey`." }}
```
egabancho, line 113: This breaks existing installs that use an external Redis without authentication. Before this PR, setting only `redisExternal.hostname` was a valid config. Now rendering fails:
```
helm template t charts/invenio -f charts/invenio/ci/minimal-values.yaml \
  --set redis.enabled=false --set redisExternal.hostname=my-redis
# Error: You have configured `.Values.redisExternal`, but no authentication credentials were provided.
```
Setting a dummy password doesn't help, because Redis rejects AUTH when no password is configured.

Also, I would drop `value:` and return the value instead like you are doing for other values.

charts/invenio/templates/_helpers.tpl
```diff
@@ -470,14 +549,7 @@ Add sentry environmental variables
 Invenio basic configuration variables
 */}}
 {{- define "invenio.configBase" -}}
-INVENIO_ACCOUNTS_SESSION_REDIS_URL: 'redis://{{ include "invenio.redis.hostname" . }}:6379/1'
 INVENIO_TRUSTED_HOSTS: '["{{ include "invenio.hostname" $ }}"]'
-INVENIO_CACHE_REDIS_HOST: '{{ include "invenio.redis.hostname" . }}'
```
egabancho, line 475: Removing these from the shared `-config` ConfigMap takes Redis settings away from every workload that doesn't also include `invenio.config.redis`. That's currently:

- `install-init-job.yaml`, which runs wipe_recreate.sh, and that flushes the cache
- `terminal-deployment.yaml`
- `worker-scheduler-deployment.yaml` (on master since cce2dcc, this branch needs a rebase 😉 )

These pods would fall back to Invenio's defaults (`localhost:6379`). Downstream charts that build their own Jobs (or any other `kind`) from the `-config` ConfigMap plus the `invenio.config.*` helpers are affected too, so this probably deserves an upgrade note.

charts/invenio/templates/web-deployment.yaml
```diff
@@ -88,6 +88,7 @@ spec:
           value: {{ required "Missing .Values.global.timezone" .Values.global.timezone }}
         {{- include "invenio.config.queue" . | nindent 8 }}
         {{- include "invenio.config.database" . | nindent 8 }}
+        {{- include "invenio.config.redis" . | nindent 8 }}
```
egabancho, line 91: Moving the Redis URLs from `envFrom` (ConfigMap) to `env` changes precedence. Kubernetes gives `env` priority over `envFrom`, so Redis overrides in `invenio.extraConfig` are now silently ignored:
```
--set-string invenio.extraConfig.INVENIO_CACHE_REDIS_URL=redis://other:6379/5
# ConfigMap:          INVENIO_CACHE_REDIS_URL: redis://other:6379/5
# web container env:  INVENIO_CACHE_REDIS_URL = $(INVENIO_CONFIG_REDIS_PROTOCOL)://...:6379/0   <- wins
```
Before this PR, `invenio.mergeConfig` let `extraConfig` replace these defaults. It's worth either keeping that working or calling out the change.

charts/invenio/values.yaml
```diff
@@ -1074,12 +1078,20 @@ redis:
...
     disableCommands: []  # Dangerous! This lets us run the `FLUSHALL` and `FLUSHDB` commands! Unfortunately, they are required by the wipe_recreate.sh script when installing Invenio.
 
-
 ## @param redisExternal External redis configuration
 ## All of these values are only used when redis.enabled is set to false
+## @extra redisExternal.password Password
 ## @extra redisExternal.hostname
+## @extra redisExternal.existingSecret Name of an existing secret resource containing the credentials
+## @extra redisExternal.existingSecretPasswordKey Name of an existing secret key containing the credentials
+## @extra redisExternal.port defaults to 6379
```
egabancho, line 1087: These are documented as configurable, but `invenio.redis.port` and `invenio.redis.protocol` in `_helpers.tpl` hardcode `6379` and `redis` and never read them. So there's no way to use a TLS-only managed Redis (`rediss://`) or a non-default port. I think we should update the helpers to use `.Values.redisExternal.port` / `.redisProtocol` with those defaults?

charts/invenio/templates/_helpers.tpl
```diff
@@ -83,10 +83,89 @@ Return the proper Invenio image name
...
+{{- end -}}
+
+{{/*
+  Redis connection env section.
+*/}}
+{{- define "invenio.config.redis" -}}
+{{- if and (not .Values.redis.enabled) (not .Values.redisExternal) }}
+  {{ fail "\n\nMissing Redis configuration.\nEnable Redis by setting `.Values.redis.enabled` to `true`, or configure `.Values.redisExternal`." }}
+{{- end }}
+{{- $uri := "$(INVENIO_CONFIG_REDIS_PROTOCOL)://:$(INVENIO_CONFIG_REDIS_PASSWORD)@$(INVENIO_CONFIG_REDIS_HOST):$(INVENIO_CONFIG_REDIS_PORT)" -}}
```
egabancho, line 141: Note that the password goes into the URL without URL-encoding. The AMQP helper already does the same, but it matters more here because Redis passwords are now user-supplied:
```
--set redis.auth.enabled=true --set 'redis.auth.password=p@ss/w:rd#1'
# INVENIO_CACHE_REDIS_URL = redis://:p@ss/w:rd#1@<host>:6379/0
```
redis-py, kombu and limits all read the host, path and fragment from that URL incorrectly, so none of them can connect. Kubernetes `$(VAR)` expansion can't encode the value, so at least a note in `values.yaml` that the password must be URL-safe would help.

egabancho (changes requested):
Thanks for picking this up, Redis auth support is a welcome addition. I left a few inline comments.

The existing unit tests pass, but none cover the new helpers. Tests for no auth, a password, existingSecret, and redisExternal with and without credentials would have caught a few of the issues and would protect this going forward. 

Happy to re-review once you're ready!

PR #254, opened 2026-08-11
feat: Enable Gateway API configuration

lindhe:
### Description

This change adds configuration parameters for `HTTPRoute` objects to work with Gateway API.

### Checklist

- [x] I've added relevant test cases.

**Reminder**


charts/invenio/templates/httproute.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  parentRefs:
+  {{- if (and .Values.listenerSet.enabled .Values.listenerSet.https) }}
+    - group: gateway.networking.k8s.io
+      kind: ListenerSet
+      name: {{ include "invenio.fullname" . }}
+      sectionName: https
+  {{- end }}
+  {{- with .Values.httpRoute.parentRefs }}
+    {{- . | toYaml | nindent 4 }}
```
egabancho, line 25:
```suggestion
    {{- tpl (. | toYaml) $ | nindent 4 }}
```
Mostly for consistency across the template.

charts/invenio/templates/listenerset.yaml
```diff
@@ -0,0 +1,34 @@
...
+    {{- if .Values.listenerSet.https }}
+    - name: https
+      hostname: {{ include "invenio.hostname" $ | quote }}
+      port: 443
+      protocol: HTTPS
+      tls:
+        certificateRefs:
+          - group: ""
+            kind: Secret
+            name: {{ printf "%s-tls" (include "invenio.fullname" .) }}
```
egabancho, line 27: The Ingress uses `invenio.tlsSecretName` instead, which gives `<hostname>-tls` or `ingress.tlsSecretNameOverride`. A deployment moving from Ingress to Gateway API would point at a secret that doesn't exist, so the HTTPS listener fails to resolve its certificate. The override would also be silently ignored.

```suggestion
            name: {{ include "invenio.tlsSecretName" $ }}
```

charts/invenio/templates/listenerset.yaml
```diff
@@ -0,0 +1,34 @@
...
+        mode: Terminate
+    {{- end }}
+  {{- with .Values.listenerSet.listeners }}
+    {{- tpl (toYaml .) $ | nindent 4 }}
+  {{- end }}
+  parentRef: {{- required "missing .Values.listenerSet.parentRef" .Values.listenerSet.parentRef | toYaml | nindent 4 }}
```
egabancho, line 33: I might be wrong here, but I think `required` will never fire for `parentRef`.
The default is `parentRef: {}`, and `required` only catches nil or an empty string, so an empty map gets through. With `listenerSet.enabled=true` and no parentRef, the chart renders `parentRef: {}` and the API server rejects it because name is missing.
Maybe we could do the check on name?
```yaml
  {{- if not .Values.listenerSet.parentRef.name }}
  {{- fail "missing .Values.listenerSet.parentRef.name" }}
  {{- end }}
```

charts/invenio/templates/listenerset.yaml
```diff
@@ -0,0 +1,34 @@
...
+    {{- . | toYaml | nindent 4 }}
+  {{- end }}
+  labels:
+    {{- include "invenio.labels" . | nindent 4 }}
+    {{- with .Values.listenerSet.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  listeners:
+    {{- if .Values.listenerSet.https }}
```
egabancho, line 18: With `listenerSet.https=false` and the default `listeners: []`, `spec.listeners` renders as null. The CRD requires at least one listener, so the install fails. A `fail` when both are empty would give a clearer error, WDYT?

charts/invenio/templates/httproute.yaml
```diff
@@ -0,0 +1,33 @@
...
+  annotations:
+    {{- . | toYaml | nindent 4 }}
+  {{- end }}
+  labels:
+    {{- include "invenio.labels" . | nindent 4 }}
+    {{- with .Values.httpRoute.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  parentRefs:
```
egabancho, line 17: Setting only `httpRoute.enabled=true`, with no `ListenerSet` and no `parentRefs`, produces a route that isn't attached to any Gateway. It installs cleanly but serves no traffic. Maybe failing at render time when neither a `ListenerSet` nor any `parentRefs` is configured, WDYT?

egabancho (changes requested):
Great work!

Just a few small comments and a request ☺️ 

Most Gateway API setups need a redirect from HTTP to HTTPS, and since the chart already sets up the HTTPS side, it would be good to support this too. Because route rules apply to every parent a route attaches to, it would have to be a separate HTTPRoute attached only to the HTTP listener. Maybe something like:

```yaml
httpRoute:
  httpsRedirect:
    ## @param httpRoute.httpsRedirect.enabled Creates an HTTPRoute that redirects HTTP to HTTPS.
    enabled: false
    ## @param httpRoute.httpsRedirect.parentRefs Templated parentRefs for the redirect route (e.g. the Gateway's HTTP listener).
    parentRefs: []
```

```yaml
{{- if and .Values.httpRoute.enabled .Values.httpRoute.httpsRedirect.enabled }}
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: {{ include "invenio.fullname" . }}-https-redirect
  labels:
    {{- include "invenio.labels" . | nindent 4 }}
spec:
  parentRefs:
  {{- if .Values.listenerSet.enabled }}
    - group: gateway.networking.k8s.io
      kind: ListenerSet
      name: {{ include "invenio.fullname" . }}
      sectionName: http
  {{- end }}
  {{- with .Values.httpRoute.httpsRedirect.parentRefs }}
    {{- tpl (toYaml .) $ | nindent 4 }}
  {{- end }}
  hostnames:
    - {{ include "invenio.hostname" $ | quote }}
  rules:
    - filters:
        - type: RequestRedirect
          requestRedirect:
            scheme: https
            statusCode: 301
{{- end }}
```

